PR #2, opened 2021-01-08
feat(api): stub CRUD for checker template

LoneRifle:
## Problem

As a pro-active effort to provide some backend facilities to the dev team,
Refactor api into its own module, then inject a controller
that wraps around a stubbed checker repository

## Solution

- Create a module that acts as a factory for the Express router
  that embodies the API, taking in an options object for dependencies
- Mount the API at `/api/v1`
- Create a CheckerController to act as the Express controller for
  `/c` endpoints.
  - Let this wrap around an object that just keys strings to
    the checker templates

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "@emotion/react": "^11.1.4",
     "@emotion/styled": "^11.0.0",
     "ace-builds": "^1.4.12",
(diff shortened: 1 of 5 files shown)


lamkeewei (approved):
lgtm!

PR #3, opened 2021-01-08
feat: add text field

alexanderleegs:
**Features**:

This PR adds an additional component for text fields. Also supports subfields for phone and email, which we implement additional regex checks for.


lamkeewei:
As discussed, let's pivot this to a date input field instead! Sorry for the change.

PR #4, opened 2021-01-11
feat: add sequelize backed by sqlite

LoneRifle:
## Problem
The current backend is not a realistic depiction of what may be the final production service

## Solution
Flesh out the backend further by introducing a Sequelize model
and necessary encapsulations to persist/retrieve checker instances.

Allow the user to specify the primary key, enforcing uniqueness. This
is in-line with how short urls are unique in GoGovSg and usable
as PKs; in a similar vein, we want urls to checkers to be memorable
to humans.

- Add sequelize, sqlite
- Rename the Config interface in types/checker to Checker
  - Add associated Sequelize model
- Create a service wrapper around Checker model, injecting all this
  into the controller
- Rewrite the controller to accommodate Sequelize (via the service)

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -2,12 +2,22 @@ import path from 'path'
 
 import express from 'express'
 import bodyParser from 'body-parser'
+import { Sequelize } from 'sequelize'
 
 import config from './config'
 import api from './api'
-import CheckerController from './checker'
+import { addModelsTo } from './models'
+import { CheckerController, CheckerService } from './checker'
 
-const checker = new CheckerController()
+const sequelize = new Sequelize({ dialect: 'sqlite' })
+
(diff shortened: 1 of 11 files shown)


src/server/index.ts
```diff
@@ -5,9 +5,16 @@ import bodyParser from 'body-parser'
 
 import config from './config'
 import api from './api'
-import CheckerController from './checker'
+import { initSequelize } from './models'
+import { CheckerController, CheckerService } from './checker'
 
-const checker = new CheckerController()
+const { models } = initSequelize()
```
lamkeewei, line 11: Should we wait for the promise returned by `sequelize.sync()` to resolve before starting the express server?

LoneRifle: Good catch! I'll change it once #10 is merged

PR #10, opened 2021-01-11
feat: add base types for checker config

lamkeewei:
Added types for Checker configuration


src/types/client/checker.d.ts
```diff
@@ -0,0 +1,59 @@
+export interface Config {
+  id: number
```
LoneRifle, line 2: Let's make this a string, letting the user specify the id. We then check in the backend whether the id is unique.

It will also be good if this interface is available to the server to, so perhaps define it at `src/types` then alias it in the client? I can then do likewise server end.

lamkeewei: @LoneRifle sorry, how do I alias it in client?

LoneRifle: maybe...
```js
export { Config } from '..'
```

LoneRifle: or alternatively, just refer to the type directly from `components/Checker`

PR #11, opened 2021-01-11
feat: set up checker evaluation framework

prestonlimlianjie:
This PR adds the ability to perform arithmetic operations using the mathjs npm package.

To test it out, go to http://localhost:3000/debug and paste the following configuration into the debugger.

```
{
	"title": "Simple a*b",
	"fields": [
    {
      "id": "IN_0",
      "type": "NUMERIC",
      "description": "Define a"
    }, {
      "id": "IN_1",
      "type": "NUMERIC",
      "description": "Define b"
	  }
  ],
	"operations": [
    {
      "id": "OUT_0",
      "type": "ARITHMETIC",
      "expression": "IN_0 * IN_1"
    },
    {
      "id": "OUT_1",
      "type": "ARITHMETIC",
      "expression": "OUT_0 + 2 * IN_1"
    }
  ],
  "displays": [
    {
      "id": "DISP_0",
      "type": "number",
      "target": "OUT_0"
    },
    {
      "id": "DISP_1",
      "type": "number",
      "target": "OUT_1"
    }
  ]
}
```


src/client/components/Checker.tsx
```diff
@@ -22,24 +53,57 @@ export const Checker: FC<CheckerProps> = ({ config }) => {
...
+        return (transformationResults[id] = TAnd(targetValues))
+      case 'IFTHEN':
+        if (args[0] === 'DISPLAYTEXT') {
+          if (targetValues[0]) {
+            setOutputString(args[1])
+          } else {
+            setOutputString(args[2])
+          }
+          return (transformationResults[id] = targetValues[0])
+        }
```
LoneRifle, line 79: the output for IFTHEN should be the strings specified as arguments depending on the expression, ie,
```suggestion
          return (transformationResults[id] = targetValues[0] ? args[1], args[2])
```

the creator then marks whichever transformation result should be displayed

prestonlimlianjie: Makes sense, change committed

src/client/transformations/index.ts
```diff
@@ -0,0 +1,30 @@
+function TNumericEqual(targets: string[], args: string[]): boolean {
```
prestonlimlianjie, line 1: Note: modify for `targets` and `args` to `number[]`

src/client/components/Checker.tsx
```diff
@@ -3,13 +3,22 @@ import { Container, Heading, Divider, Stack, Button } from '@chakra-ui/react'
...
 interface CheckerProps {
   config: any
 }
 
+interface Operation {
+  id: string
+  type: string
+  expression: string
+}
+
```
LoneRifle, line 17: Eh, rebase your codebase so that you can pick up some of the latest changes (including the declaration of the Checker type, which should take on board this interface)

src/client/components/Checker.tsx
```diff
@@ -22,17 +31,35 @@ export const Checker: FC<CheckerProps> = ({ config }) => {
...
+      // case 'SWITCH':
+    }
+  }
+
   // TODO: For checkbox and radio since the value is the option index, we need to retrieve the actual value.
-  const onSubmit = (data: any) => console.log(data)
+  const onSubmit = (data: any) => {
+    // Define the input data variables
+    variables = data
+    operations.map(evaluateOperation)
```
lamkeewei, line 51: Maybe declare variables with `useState` and update it this way:

```suggestion
    const variables = operations.reduce(evaluateOperation, data)
    setVariables(variables)
```

Issue with the current way of declaring `variables` is that it gets reinitialised every time there's a re-render and we will lose the state. Will be an issue when we update the UI to display the results later on.

PR #12, opened 2021-01-11
Feat/date field

alexanderleegs:
Features:

This PR adds an additional component for date fields. As Chakra UI does not natively support date inputs, we added the react-datepicker library to use instead, and use a hidden input field to store the data.

This PR also updates our webpack config to read css files.

### DIFF
--- a/src/client/components/fields/DateField.tsx
+++ b/src/client/components/fields/DateField.tsx
@@ -44,6 +44,9 @@ export const DateField: FC<DateProps> = ({ order, id, description, help }) => {
         selected={date}
         onChange={changeHandler}
         showPopperArrow={true}
+        showMonthDropdown
+        showYearDropdown
+        dropdownMode="select"
       />
       <FormErrorMessage>Field is required</FormErrorMessage>
     </FormControl>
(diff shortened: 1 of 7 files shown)


src/client/components/fields/DateField.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return (
+    <FormControl isInvalid={error} id={id}>
+      <FormLabel htmlFor={id}>{`${order + 1}. ${description}`}</FormLabel>
+      {help && <FormHelperText mb={4}>{help}</FormHelperText>}
+      <Input
+        type="hidden"
+        name={id}
+        value={date && date.toISOString().split('T')[0]}
```
lamkeewei, line 39: Since variables will be passed into the final formula through string substitution, might be easier if we represent date as milliseconds since epoch. Avoid having to parse the date format again when evaluating for the result.

alexanderleegs: Good point, updated in https://github.com/opengovsg/checkfirst/pull/12/commits/1d65016a3ee19d586db3bfe0d06c5f3426cecb6e

lamkeewei:
Just thinking, is there a quick way to select year? Currently, if the field is for something like birth date, a user will have to scroll for quite some time.

alexanderleegs:
> Just thinking, is there a quick way to select year? Currently, if the field is for something like birth date, a user will have to scroll for quite some time.

react-datepicker has a dropdown for year and month, I've added this in https://github.com/opengovsg/checkfirst/pull/12/commits/9d602dd42b25a3730a1e917f35acab89753ec67e

PR #20, opened 2021-01-14
feat: ifelse and string equality comparison

lamkeewei:
## Solution

**Features**:

- Added `ifelse` math.js function
- Added support for string equality comparison

**Improvements**:
- Added `ts-jest` for testing

## Tests
- `npm test`


src/client/core/evaluator.ts
```diff
@@ -0,0 +1,50 @@
...
+      return a === b
+    }
+  ),
+  createUnequal: factory(
+    'unequal',
+    [],
+    () =>
+      function unequal(a: string | number, b: string | number) {
+        return a !== b
+      }
```
LoneRifle, line 21: nitpick - arrow function for consistency with lines 11-13?
```suggestion
    () => (a: string | number, b: string | number) => {
      return a !== b
    }
```

src/client/core/evaluator.ts
```diff
@@ -0,0 +1,50 @@
...
+        condition: boolean,
+        a: string | number,
+        b: string | number
+      ) => (condition ? a : b),
+    })
+  ),
+}
+export const math = create(factories, config)
+
+export function Operation(
```
LoneRifle, line 36: nitpick - it's not a class name, so..
```suggestion
export function operation(
```

src/client/core/evaluator.ts
```diff
@@ -0,0 +1,50 @@
...
+  expression: string,
+  variables: Record<string, string | number>
+): number {
+  return math.evaluate!(expression, variables)
```
LoneRifle, line 40: why is the non-null assertion needed here?

LoneRifle (changes requested):
Just a bunch of nitpicks, one question

package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "lint": "eslint .",
     "lint-fix": "eslint --fix .",
-    "test": "echo 'No tests implemented'",
+    "test": "jest --watch",
```
LoneRifle, line 10: This really should be...
```suggestion
    "test": "jest",
```

if you need to watch you could do...
```
npm test -- --watch
```

src/client/core/evaluator.ts
lamkeewei: I'm stumped by his too. Suspect that it's something to do with type definitions for math.js. But haven't looked deeply into it.

LoneRifle (approved):

src/client/core/evaluator.ts
```diff
@@ -0,0 +1,49 @@
...
+export const math = create(factories, config)
+
+export function operation(
+  expression: string,
+  variables: Record<string, string | number>
+): number {
+  return math.evaluate!(expression, variables)
+}
+
+export const evaluateOperation = (
```
prestonlimlianjie, line 42: Could we rename this to `operationReducer`/`variableReducer` or something similar to capture the idea that it is a reducer function that produces a final `variables` object?

src/client/core/evaluator.ts
```diff
@@ -0,0 +1,49 @@
...
+        condition: boolean,
+        a: string | number,
+        b: string | number
+      ) => (condition ? a : b),
+    })
+  ),
+}
+export const math = create(factories, config)
+
+export function operation(
```
prestonlimlianjie, line 35: Could we rename this function to `evaluateOperation` because `operation` is a little misleading. This function uses the expression in the `Operation` object and proceeds to evaluate it

LoneRifle: Sounds fair - maybe @alexanderleegs can follow up with a PR

prestonlimlianjie: I'll make the PR

PR #21, opened 2021-01-14
style: rename evaluation functions

prestonlimlianjie:
This PR renames two evaluator functions described in https://github.com/opengovsg/checkfirst/pull/20#discussion_r557253157 and https://github.com/opengovsg/checkfirst/pull/20#discussion_r557250266 for clarity.

### DIFF
--- a/src/client/components/Checker.tsx
+++ b/src/client/components/Checker.tsx
@@ -4,7 +4,7 @@ import { useForm, FormProvider } from 'react-hook-form'
 
 import { Checkbox, Radio, Numeric, DateField } from './fields'
 import * as checker from './../../types/checker'
-import { evaluateOperation } from './../core/evaluator'
+import { variableReducer } from './../core/evaluator'
 
 interface CheckerProps {
   config: checker.Checker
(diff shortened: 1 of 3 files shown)


LoneRifle:
Ah, I should have spotted this earlier - function names should be verbs. Could I suggest `evaluate` and `reduce`? These are more concise and the nouns they are acting on are already implied in the args

src/client/core/evaluator.ts
```diff
@@ -32,18 +32,18 @@ const factories = {
 }
 export const math = create(factories, config)
 
-export function operation(
+export function evaluateOperation(
```
LoneRifle, line 35: Consider the more concise...
```suggestion
export function evaluate(
```

src/client/components/Checker.tsx
```diff
@@ -29,10 +29,7 @@ export const Checker: FC<CheckerProps> = ({ config }) => {
   }
 
   const onSubmit = (inputVariables: Record<string, string | number>) => {
-    const computedVariables = operations.reduce(
-      evaluateOperation,
-      inputVariables
-    )
+    const computedVariables = operations.reduce(variableReducer, inputVariables)
```
LoneRifle, line 32: A bit conflicted here - function names should be verbs but given you're mostly using this as a lambda here, a noun for a name is somewhat suitable. Leaving this to you.

LoneRifle (approved):
Lgtm otherwise

src/client/components/Checker.tsx
prestonlimlianjie: Referring to MDN, where they call it a `reducer`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce

As much as I dislike it as well, it might be clearer to follow their convention

PR #30, opened 2021-01-18
feat: login page and frontend authentication

lamkeewei:
## Solution

Add login page and frontend authentication. 

## Tests
- Login with a valid email
- Login without a valid email
- Resend OTP
- `/projects` should only be available after login

**New dependencies**:
- `react-query` : Hooks for fetching, caching and updating asynchronous data in React
- `axios` : Library for making HTTP calls

## TODOs
- Need `/logout` route
- Need `/user/:id` route to check if user is already logged in
- Use design from Natalie


LoneRifle (approved):
lgtm

PR #37, opened 2021-01-20
refactor(login): use Sign In/Out for text, react-icon

LoneRifle:
Line up the interface to be closer to the wireframes by:
- using the appropriate text in the landing page, and;
- using `<BiLogOutCircle />` for the icon


lamkeewei (approved):
lgtm

PR #38, opened 2021-01-20
feat(projects): implement navbar

LoneRifle:
## Context
image

The projects page has to be fleshed out to allow users to create and manage checkers. 

This PR seeks to put in place a navigation bar. It also changes AuthContext to expose
more information, information that the user will expect to see on successful login.

## Details

- Copy the navbar from builder, and restructure it to fit wireframe
- Create a common LogoutButton and use in projects and builder navbars,
  complete with functionality to logout and redirect to home page
- Deliberately avoid the tab interface proposed for the dashboard,
  as we are unlikely to have settings, leaving just one panel


lamkeewei (approved):
lgtm!

PR #41, opened 2021-01-20
refactor(builder): use proper tabs

LoneRifle:
https://user-images.githubusercontent.com/10572368/105133635-0e336200-5b28-11eb-892b-0b67380d4b36.mov

## Context 

Implement proper tabbing in FormBuilder, so that have a proper
place to put both layout and logic UI elements

- Introduce Tabs in Builder.Navbar and TabPanels in FormBuilder
- Wire the two together using a state hook


lamkeewei (approved):
lgtm!

PR #45, opened 2021-01-20
fix(checker): link Checkers and Users, create Checker under User

LoneRifle:
## Problem and Solution

- Ensure that Checkers and Users are related on many-to-many basis
- CheckerController: ensure that user has a session before creation
- CheckerService: Lookup user before creating checker, and link
  via implicit join table
- Add constants to Checker model
- Reorder fields in checker type


lamkeewei (approved):
lgtm!

PR #62, opened 2021-01-22
feat(checker): authenticate user before operations

LoneRifle:
## Problem

Checkers are freely accessible by all because there is no user authorization checks actually being done

## Solution

- Ensure the user can update/delete the checker
- Ensure the user lists only the checkers he/she has access to
- When retrieving a checker, if anonymous or user with no access to
  checker, return the checker without user information
- Similarly, if user is retrieving his/her own checker, add user
  information

### DIFF
--- a/src/server/checker/CheckerController.ts
+++ b/src/server/checker/CheckerController.ts
@@ -28,19 +28,25 @@ export class CheckerController {
     }
   }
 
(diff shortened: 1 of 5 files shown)


lamkeewei (approved):
lgtm!

PR #64, opened 2021-01-24
feat: validate and derive evaluation order for operations with topological sort

lamkeewei:
## Problem
Given that operations can depend on the outputs of each other, we need a way to:
- Determine the right order of evaluation. This means that for a given operation all of its dependencies should be evaluated before evaluating itself
- Detect possible cyclical dependency chains. E.g. if `A -> B`, `B -> C` and `C -> A`, we should be able to detect the cycle and throw an appropriate error

## Solution
Valid checker logic can be represented as a directed acyclic graph (DAG). As such, we can exploit the following properties:
- Linear ordering of a DAG can be obtained through a [topological sort](https://en.wikipedia.org/wiki/Topological_sorting)
- A valid DAG has at least one topological ordering. Otherwise, there is a cycle in the graph and it is not a DAG

As such, I'm proposing the following solution:
- On every evaluation, construct a directed graph from the inputs, constants and operations
- During graph construction check that each operation's dependencies are valid (they should exist in the graph)
- Carry out a topological sort to obtain the order of evaluation
- Run the evaluation of the operations based on the derived order

In addition, `getEvaluationOrder` can also be used as a function to validate a set of inputs, constants and operations.

This static analysis of the computation graph has the following added advantages over a dynamic analysis (running evaluate for each operation linearly with dummy values): 
- No need to infer the type of the operation's dependencies and figure out what appropriate dummy values to use
- No need to force the user to arrange the operations in a specific order in the UI since we are sorting on evaluation
- By detecting cycles, we can pinpoint where the circular dependencies are and reflect that in the UI

## Tests
- Run unit tests

## Deploy Notes
**New dependencies**
- `graphlib` - Library that provides graph operations and algorithm (we're using `toposort` and `findCycles`)


src/client/core/evaluator.ts
```diff
@@ -47,3 +60,113 @@ export const variableReducer = (
...
+export const getDependencies = (expression: string): Set<string> => {
+  // Parse the mathjs expression into an expression tree
+  const node = math.parse!(expression)
+
+  // Traverse tree to extract symbols and functions
+  const dependencies = new Set<string>()
+  const functions: string[] = []
+  node.traverse((node) => {
+    if (node.isFunctionNode && node.name) functions.push(node.name)
+    if (node.isSymbolNode && node.name) dependencies.add(node.name)
```
prestonlimlianjie, line 77: could we do `node.isSymbolNode && !node.isFunctionNode && node.name` instead?

prestonlimlianjie (approved):
An elegant solution! lgtm with minor comment

src/client/core/evaluator.ts
lamkeewei: Wanted to do that at first. But the way the expression tree is structure for say `ifelse(A == B, C, D)` is that there will be two nodes for `ifelse`, one for a `SymbolNode` and one for a `FunctionNode`.

PR #66, opened 2021-01-25
feat(production): prepare for deployment

LoneRifle:
## Problem

The application currently is only fit for running on a development machine, and is not ready for deployment to production. Notable missing features include the lack of proper logging and the inability to connect to a database and SMTP server for mailing OTPs.

## Solution

- Move most of the bootstrapping block in index.ts into its own module, 
  dividing the block into specific modules for sequelize, nodemailer and logging
- Allow nodeEnv to influence what the modules expose to bootstrap,
  and create sensible objects to use if nodeEnv is not development
- Take the trouble to tweak how Nodemailer's `sendMail()` is injected
  into AuthService, on learning that it can return a Promise
- Add morgan logging to express bootstrapping
- Accept logger as an optional dependency in AuthController
  and CheckerService, logging useful messsages in lieu of console.{error, warn}
- Take the trouble to log unauthorized OTP attempts
- Expose the Express app as a AWS Lambda through serverless-http

**Notes**
Because we have to wait for `sequelize.sync()` before making the
Express app available, we cannot obtain a serverless handler directly
if we are going to use it in top-level module code, and have to access
it through a Promise. However, AWS requires us to expose the serverless
handler alone, and cannot unpack it from a Promise resolving to one.

To get around this, implement an AWS Lambda handler that will await
on the Promise that is creating the serverless handler from the
app returned by the asynchronous `bootstrap()`. In effect, we are
unwrapping the Promise on AWS' behalf.

**TODO**
Configure for use on AWS Lambda + API Gateway using the serverless framework

- `DATABASE_URL` : a URL to inject into Sequelize
- `SMTP_URL`: a URL to inject into nodemailer

**New dependencies**:
- `morgan`, `winston` - logging dependencies
- `serverless-http` - an adapter for Connect/Express to AWS Lambda


src/server/bootstrap/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  const apiMiddleware = [sessionMiddleware, bodyParser.json()]
+  app.use('/api/v1', apiMiddleware, api({ checker, auth }))
+
+  // Facilitate deep-linking
+  app.get('*', (_req: Request, res: Response) => {
+    res.sendFile(path.resolve(__dirname + '/../../build/client/index.html'))
+  })
+
+  await sequelize.sync()
```
lamkeewei, line 102: Would this introduce significant overheads for each lambda startup? But I guess eventually we can handle database migrations separately rather than depending on the `sync` function for prod.

lamkeewei (approved):
lgtm, with a minor question on `sequelize.sync`.

prestonlimlianjie (approved):
lgtm

src/server/bootstrap/index.ts
LoneRifle: Maybe not for the lambda startup, but this might pose a challenge to the underlying database which might have to deal with an influx of spurious sync requests. We can revisit this once we are more established.

LoneRifle:
Followed up with 651a776 to fix some problems. Notably:

- added `pg` as a npm dep
- allowed `DATABASE_URL` and `SMTP_URL` to be picked up and acted on even if `NODE_ENV` is development

PR #72, opened 2021-01-26
feat: render displays using operations

prestonlimlianjie:
## Feature

This PR adds the following features:
- allows admin users to hide or show the various Logic components
- renders `LineDisplay` using the `Operations` array


lamkeewei (approved):
lgtm!

PR #73, opened 2021-01-26
fix: generate unique id for each field and operation

alexanderleegs:
This PR introduces a simple running sequence of numbers to generate unique ids for fields and operations.

### DIFF
--- a/src/client/components/builder/QuestionsTab.tsx
+++ b/src/client/components/builder/QuestionsTab.tsx
@@ -24,16 +24,16 @@ import { BuilderActionEnum, ConfigArrayEnum } from '../../../util/enums'
 
 const TITLE_FIELD_INDEX = -1
 
-const defaultNumericField: checker.Field = {
-  id: 'N',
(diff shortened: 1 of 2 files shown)


src/client/components/builder/QuestionsTab.tsx
```diff
@@ -24,41 +24,42 @@ import { BuilderActionEnum, ConfigArrayEnum } from '../../../util/enums'
 
 const TITLE_FIELD_INDEX = -1
 
-const defaultNumericField: checker.Field = {
-  id: 'N',
+const generateDefaultNumericField = (id: number): checker.Field => ({
+  id: `N${id}`,
```
prestonlimlianjie, line 28: The PR looks good - could we check with Nat if she's okay with these identifiers (`N`, `R`, etc)?

alexanderleegs: The identifiers used are okay - just need to make changes if any field requires the same letter in the future

PR #74, opened 2021-01-26
feat(builder): warn user with modal if unsaved changes

LoneRifle:
https://user-images.githubusercontent.com/10572368/105811464-a4173300-5fe7-11eb-907f-97c8d392df68.mov

## Problem

We currently do not prompt when the user has unsaved changes 

## Solution

- When loading checker from backend, keep a copy of that config
- Compare copy against current state of checker config, and mark
  if there are changes
- Use that to enable/disable the save button
- If `isChanged`, warn the user about discarding changes, navigating
  back to `/dashboard` on confirmation
- Change title placeholder in navbar from button to text


lamkeewei (approved):
lgtm!

PR #78, opened 2021-01-26
fix: prevent restart of field numbering when reloading page

alexanderleegs:
This PR fixes a bug where the numbering of fields and operations would restart upon reloading the page. We now look through the existing items in fields/operations to determine the appropriate index to use.

This PR also fixes a minor bug involving the reordering of operations when moving down.

### DIFF
--- a/src/client/components/builder/LogicTab.tsx
+++ b/src/client/components/builder/LogicTab.tsx
@@ -41,8 +41,8 @@ export const LogicTab: FC = () => {
 
   useEffect(() => {
     let highestIndex = 0
-    config.operations.forEach((field) => {
-      const operationIndex = parseInt(field.id.slice(1))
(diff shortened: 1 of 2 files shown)


src/client/components/builder/LogicTab.tsx
```diff
@@ -39,6 +39,15 @@ export const LogicTab: FC = () => {
   const [nextUniqueId, setNextUniqueId] = useState<number>(1)
   const { dispatch, config } = useCheckerContext()
 
+  useEffect(() => {
+    let highestIndex = 0
+    config.operations.forEach((field) => {
```
prestonlimlianjie, line 44: Could we use `operation` as the param since this is `forEach()` over `operations`?

prestonlimlianjie (approved):
1 small change request but apart from that, lgtm

PR #79, opened 2021-01-27
feat(builder): implement embed modal

LoneRifle:
image

## Problem
The user does not necessarily know how to share the checker or embed it in a government website

## Solution

- use a second set of disclosure hooks for a second modal, renaming
  the first set to a more relevant name for prompting when clicking
  back
- implement the embed modal, allowing iframe and link to be copied
  to clipboard with one click


lamkeewei (approved):
lgtm!

PR #81, opened 2021-01-27
feat: add ConditionalResult component

lamkeewei:
## Solution
- Added conditional result component.
- Expression is still stored as a string at the top level CheckerContext. When rendering, string expression is parsed into `conditions`, `ifExpr`, `thenExpr` and `elseExpr`. 

## Screenshot
image

## TODOs
- Make parsing more robust
- Fix some styling issues


src/client/components/builder/logic/ConditionalResult.tsx
```diff
@@ -0,0 +1,289 @@
...
+}
+
+const EMPTY_STATE: IfelseState = {
+  ifExpr: '',
+  conditions: [],
+  elseExpr: '',
+  thenExpr: '',
+}
+
+const fromExpression = (expression: string): IfelseState => {
```
lamkeewei, line 42: I guess a more robust way will be to use `mathjs` to parse this and then traverse the expression tree to build the string. Will probably get back to that in the future.

PR #86, opened 2021-01-27
fix: return option labels instead of values for Radio and Checkbox fields

prestonlimlianjie:
This PR applies a transformation step to the field input values for the Radio and Checkbox fields.

**Radio**
Currently, the input value of the radio field is the `value` (index) of the selected option, e.g. `"0"`. The new change applies a transformation to set the input value as the `label` of the selected option label, e.g. `"Option A"`.

**Checkbox**
Currently, the input value of the checkbox field is an array of selected option `value`, e.g. `["0", "1"]`. The new change a transformation to set the input value as a serialized string representing the array of selected option labels, e.g. `"[\"Option A\",\"Option B\"]"`

### DIFF
--- a/src/client/components/Checker.tsx
+++ b/src/client/components/Checker.tsx
@@ -57,18 +57,28 @@ export const Checker: FC<CheckerProps> = ({ config }) => {
     )
   }
 
-  // An input variable is a numeric field if its key name starts with N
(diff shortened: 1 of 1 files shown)


lamkeewei (approved):
lgtm!

PR #87, opened 2021-01-28
feat: added ExpressionInput

lamkeewei:
## Solution

image

## Known issues
- If you have more than 1 `@` in the line, it will not trigger replacement (since it doesn't know what to replace)
- Inserting a new variable through the autocomplete at the start will clear the rest of the line if it's not empty

## Future extensions
- Allow suggestion for function names with help


LoneRifle (approved):
lgtm, merge at your leisure or wait for Preston

PR #88, opened 2021-01-28
feat(sqlite): persist to file

LoneRifle:
## Problem

When run on local machine, sqlite needs to persist data between runs

## Solution

By default, persist sqlite data to disk, path specified by
`SQLITE_PATH`. If that path is an empty string, assume storage
is in-memory

- Declare a new env var, `SQLITE_PATH`, default `/tmp/checkfirst.db`
- When creating a Sequelize instance, for the sqlite dialect, if
  an empty `SQLITE_PATH` is given, assume that in-memory storage
  is desired (ie, the sequelize default)
- Because sqlite does not handle transaction locking very well (eg,
  a txn cannot write to a table if it the same txn has locked it for
  reading), detect if sequelize is using sqlite backed by a file,
  and avoid using a transaction if so

- Take the trouble to extend cookie age to 4 hours

- `SQLITE_PATH` : the file to persist sqlite data to, or blank if in-memory preferred

### DIFF
--- a/src/server/bootstrap/sequelize.ts
+++ b/src/server/bootstrap/sequelize.ts
@@ -4,12 +4,16 @@ import logger from './logger'
 import config from '../config'
 
 const databaseUrl = config.get('databaseUrl')
+const sqlitePath = config.get('sqlitePath')
 
 export const sequelize = databaseUrl
   ? new Sequelize(databaseUrl, {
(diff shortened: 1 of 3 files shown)


prestonlimlianjie (approved):
lgtm

PR #147, opened 2021-02-18
feat: reset empty inputs to default values and add checker backend validation

### LINKED ISSUE #116: Validation for questions tab
There's currently no validation for user input on the questions tab. User are able to create form with empty fields.

Add necessary validation for form fields.

lamkeewei:
## Problem

There was no validation on the different fields in the question tab of the builder. It was possible to save an empty field or a radio question with no option.

Closes #116 

## Solution

**Features**:
- For all compulsory input fields (e.g. title, question title, etc), if the value is empty on blur, the value is reset to a default value. This is similar to Google Forms. This method was chosen over preventing the user from switching away from an active field if there is an invalid input as it is more seamless. 
- Prevent deletion of the final option for radio and checkbox questions.
- Add validation using Joi for checker in backend for `POST` and `PUT` endpoints

## Tests
**Reset empty inputs**
- Attempt to set each of the following as empty:
  - Title input
  - Question title input
  - Question option input
- The inputs should be reset to a default value on blur

**Deletion of only option**
- For radio and checkbox question field, deletion should be disabled if there is only one option left

## TODOs
- [x] Validate the entire fields property of a checker on each save in the backend


lamkeewei:
Closing this to separate backend validation and frontend validation into separate PRs

PR #155, opened 2021-02-22
feat: add constants tab

prestonlimlianjie:
## Overview

This PR adds the Constants Tab and allows admin users to create map tables in the tab. An admin user can now do a map lookup using a calculated field with the following syntax:`T1["field"]`, assuming that `T1 = { "field": 0 }`.

The Logic Tab map table component will be added in https://github.com/opengovsg/checkfirst/pull/178 to keep the code review manageable.

### Using an array to store the key-value pairs for the `table` (map) instead of an object
We initially considered defining the `Constant` interface with the following properties:
```
export interface Constant {
  id: string
  title: string
  table: Record<string, number>
}
```

However, defining `table` as a Record made the process of updating the key-value pair in the map table extremely challenging as the `key` is mutable. 

Consider the following case - we have a map table containing only 1 key-value pair: `{ age: 20 }`. The admin user now wants to rename the key `age` to `currentAge`. In order to do so, the `ConstantsTab` needs to be able to track both the previous and current value of the key so that it can i) delete the previous key and ii) assign the current value to the new key name. This process is challenging and requires the use of a `ref`, which adds complexity to the code and is un-React-like.

Instead, we define `table` as an array:
```
export interface TableElem {
  key: string
  value: number
}

export interface Constant {
  id: string
  title: string
  table: TableElem[]
}
```

As such, we can easily perform the update step by replacing the relevant text corresponding to the array index. With the same example above, the `ConstantsTab` just needs to modify the `key` in the 0-indexed element in the `table` array. This pattern should be familiar because we use this same pattern to update the options in the `CheckboxField` component.

### Using `NaN` to represent an `undefined` value in new table rows

This issue is a little janky. Please feel free to suggest edits.

An admin user is able to add new key-value pairs to a map table by clicking on the plus icon in the table. See below for a visual explanation.

Screenshot 2021-02-24 at 4 33 59 PM
Screenshot 2021-02-24 at 4 34 04 PM

In order to create these two inputs for the key-value pair, we push a new table element to the existing table with the following default values:
```
{ key: "", value: NaN }
```

We can represent the empty key with an empty string `""` - however, we cannot do so for the value because it is defined as type number. `null` and `undefined` are also not accepted by the compiler because they are not type number. Instead, we set the value as `NaN` and suppress its output in:
- the active map table by setting input as type `number` - [source](https://github.com/opengovsg/checkfirst/pull/155/files#diff-6a630b9620d434c088a6943eefc63145c6a060c93ddd2b81526ffd041fb669f3R113)
- the inactive map table by displaying an empty `<td />` if the value is `NaN` - [source](https://github.com/opengovsg/checkfirst/pull/155/files#diff-6a630b9620d434c088a6943eefc63145c6a060c93ddd2b81526ffd041fb669f3R174)


src/client/components/builder/ConstantsTab.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const onActive = ({ top }: { top: number }) => {
+    setOffsetTop(top)
+  }
+
+  const addInitialMap = () => {
+    dispatch({
+      type: BuilderActionEnum.Add,
+      payload: {
+        element: generateDefaultMap(0),
```
lamkeewei:
```suggestion
        element: generateDefaultMap(nextUniqueId),
```

Currently, the first constant has an ID of `T0`. We should probably standardise the numbering to start from 1, similar to questions and logic.

src/client/components/builder/ConstantsTab.tsx
```diff
@@ -0,0 +1,158 @@
...
+      onSelect,
+      setActiveIndex,
+      nextUniqueId,
+      setNextUniqueId,
+    }
+    return <MapTable {...commonProps} />
+  }
+
+  return (
+    <VStack align="stretch" position="relative" spacing={4}>
```
lamkeewei, line 136:
```suggestion
    <>
        <Alert status="info">Constants can only be numeric values...</Alert>
        <VStack align="stretch" position="relative" spacing={4}>
```

We should add the info block to give some context to the constants tab.

image

src/client/components/builder/constants/MapTable.tsx
```diff
@@ -0,0 +1,189 @@
...
+            <Thead>
+              <Tr>
+                <Th>Reference</Th>
+                <Th>Constant</Th>
+                <Th />
+              </Tr>
+            </Thead>
+            <Tbody>
+              {table.map(({ key, value }, index) => (
+                <Tr key={index}>
```
lamkeewei, line 174:
```suggestion
                key && <Tr key={index}>
```

Should we skip rendering the entire row if key is also empty?

src/client/core/evaluator.ts
```diff
@@ -189,8 +189,16 @@ export const evaluate = (
...
+  variables = constants.reduce((vars, { id, table }) => {
+    // Convert table array to record/object
+    const tableObj = table.reduce((obj, { key, value }) => {
+      obj[key] = value
+      return obj
+    }, <Record<string, number>>{})
+
+    // Evaluate stringified object - the only way to
+    // obtain a non-escaped string of the object.
+    vars[id] = math.evaluate!(JSON.stringify(tableObj))
```
lamkeewei:
```suggestion
    vars[id] = tableObj
```

Just wondering, can we directly set the mapping table this way? https://runkit.com/embed/4jl44k2tcjtv

src/client/components/builder/ConstantsTab.tsx
prestonlimlianjie: addressed with cb2795f

src/client/components/builder/ConstantsTab.tsx
prestonlimlianjie: addressed with e243cb3

src/client/components/builder/constants/MapTable.tsx
prestonlimlianjie: I think we should still render it, because as of now, users can fill in only the `value` but not the `key`. And they might not realize what the issue is if we throw an error but hide the missing row:
Screenshot 2021-02-26 at 5 28 39 PM
Screenshot 2021-02-26 at 5 28 43 PM

src/client/core/evaluator.ts
prestonlimlianjie: This is a great question. I was playing around with 2 options before settling on the commit:

### Option 1: Setting the value as a Record<string, number>
```
vars[id] = tableObj
```

In order to get this to work, we have to make modifications to several functions to get them to return `Record<string, string | number | Record<string, number>>` instead of `Record<string, string | number>`. 

However, even after doing so, we get this compiler error.
Screenshot 2021-02-26 at 5 58 49 PM

We take a look at the exact code segment for more details on the error. It appears that a possible type of `computed` is `Record<string, ReactText>`.

Screenshot 2021-02-26 at 6 03 07 PM

And `computed` is defined as the return value of `evaluate()`.

Screenshot 2021-02-26 at 6 04 54 PM

This happens despite already having coerced the return type of `evaluate()` to be `Record<string, string | number | Record<string, number>>`.

Screenshot 2021-02-26 at 6 02 43 PM

This is quite strange and I'm not sure how to fix it. Do let me know if you have any suggestions.


### Option 2: Setting the value as a stringified Record<string, number>
```
vars[id] = JSON.stringify(tableObj)
```

However, this resulted in the variable value being escaped, and so mathjs isn't able to properly handle it. Refer to `T0`.
Screenshot 2021-02-26 at 5 48 47 PM

### Solution: evaluate a stringified Record<string, number>
```
vars[id] = math.evaluate!(JSON.stringify(tableObj))
```

This produces a value that mathjs is able to use. Refer to `T0`.

Screenshot 2021-02-26 at 5 46 51 PM

src/client/components/builder/ConstantsTab.tsx
prestonlimlianjie: Screenshot 2021-02-26 at 6 08 35 PM

src/client/core/evaluator.ts
lamkeewei: I managed to get it to compile and it seemed to work with the changes [here](https://github.com/opengovsg/checkfirst/compare/feat/add-constants...feat/add-constants-keewei).

prestonlimlianjie: Merged your proposed changes #189

lamkeewei (approved):
lgtm!

PR #184, opened 2021-02-25
feat: add date field

barnettx:
## Problem

Users don't have a way to add/subtract dates. eg. Calculating 14 day stay home notice

### Date field in questions
image

### Date result in logic
image

image

Empty date result
image

### Date field in checker
image

image


barnettx:
## UI Fixes

Reduce opacity of icon to match disabled appearance
image

Skin date picker to look more like Chakra (without borders, add shadow)
Add date icon in the input box
image

src/client/components/Checker.tsx
```diff
@@ -17,12 +17,17 @@ import { CheckboxField, RadioField, NumericField, DateField } from './fields'
...
+import { unit, Unit } from 'mathjs'
 
 interface CheckerProps {
   config: checker.Checker
 }
 
-type VariableResults = Record<string, string | number>
+type VariableResults = Record<string, string | number | Unit>
+
+function isUnit(output: string | number | Unit): output is Unit {
```
barnettx: might refactor this out?

src/client/components/builder/logic/DateResult.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+const EMPTY_STATE: DateState = {
+  variableId: '',
+  isAdd: true,
+  numberOfIntervals: 0,
+}
+
+// TODO: Make more rigorous to include other time units
+const fromExpression = (expression: string): DateState => {
+  const splitExpression = expression.split(' ')
```
barnettx, line 39: Opted to just split the expression string rather than parsing like in `ConditionalResult`

src/client/styles/date-picker.css
```diff
@@ -1,6 +1,8 @@
 .react-datepicker {
   font-family: unset;
   font-size: 0.9rem;
+  border: none;
+  box-shadow: rgb(0 0 0 / 10%) 0px 4px 6px -1px, rgb(0 0 0 / 6%) 0px 2px 4px -1px;
```
barnettx: used a chakra box-shadow. not sure if there's a better way to interface with Chakra?

src/client/core/evaluator.ts
```diff
@@ -106,7 +106,8 @@ export const getDependencies = (expression: string): Set<string> => {
   const functions: string[] = []
   node.traverse((node) => {
     if (node.isFunctionNode && node.name) functions.push(node.name)
-    if (node.isSymbolNode && node.name) dependencies.add(node.name)
+    if (node.isSymbolNode && node.name && !['day', 'days'].includes(node.name))
```
LoneRifle: As discussed... you may have to tweak this suggestion further to address linting issues
```suggestion
    if (node.isSymbolNode && node.name && !Object.keys(math.Unit.UNITS).includes(node.name))
```

src/client/components/Checker.tsx
lamkeewei: This is refactored out in Preston's latest PR. Can you rebase on `develop` and merge in the changes?

src/client/components/fields/DateField.tsx
```diff
@@ -8,46 +8,64 @@ import {
...
-        selected={date}
-        onChange={changeHandler}
-        showPopperArrow={true}
-        showMonthDropdown
-        showYearDropdown
-        dropdownMode="select"
-      />
-      <FormErrorMessage>Field is required</FormErrorMessage>
-    </FormControl>
+    <Controller
```
lamkeewei, line 26: Validation for the `DateField` doesn't seem to be triggering. Error is only thrown at run time. 

image

src/client/components/fields/DateField.tsx
```diff
@@ -8,46 +8,64 @@ import {
...
+}
+
+const DatePickerInput: FC<DatePickerInputProps> = ({ value, onClick }) => (
+  <InputGroup>
+    <Input
+      type="text"
+      placeholder="DD/MM/YYYY"
+      className="react-datepicker-ignore-onclickoutside"
+      value={value}
+      onClick={onClick}
```
lamkeewei:
```suggestion
      onClick={onClick}
      readOnly
```

Add to suppress warning:
```
You provided a `value` prop to a form field without an `onChange` ... Otherwise, set either `onChange` or `readOnly`
```

src/client/components/fields/DateField.tsx
```diff
@@ -8,46 +8,64 @@ import {
...
 
 import { Field } from '../../../types/checker'
+import { BiCalendar } from 'react-icons/bi'
+
+interface DatePickerInputProps {
+  value: string
+  onClick: (event: React.MouseEvent<HTMLInputElement, MouseEvent>) => void
+}
+
+const DatePickerInput: FC<DatePickerInputProps> = ({ value, onClick }) => (
```
lamkeewei: Might want to look into the following warning:
```
Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

src/client/components/fields/DateField.tsx
```diff
@@ -8,46 +8,64 @@ import {
...
+    <Controller
+      name={id}
+      control={control}
+      render={({ onChange, value }, { invalid }) => (
+        <FormControl isInvalid={invalid} id={id}>
+          <FormLabel sx={styles.label} htmlFor={id}>
+            {title}
+          </FormLabel>
+          {description && <FormHelperText mb={4}>{description}</FormHelperText>}
+          <DatePicker
```
lamkeewei: Was wondering if we are able to implement the multiple level of selection in the design from Nat. From what I gather from the `react-datepicker` docs, perhaps we can use `renderCustomHeader` + `showMonthYearPicker` + `showYearPicker` to achieve it.

lamkeewei (commented):
Some minor comments on the date picker component and warnings. Also rebase on develop. Thanks!

barnettx:
Updated UI using react-datepicker

image

image

image

Mobile on my Android

image

image

image

On mobile, there doesn't seem to be an easy way to change the date formatting from dd/MM/yyyy (https://stackoverflow.com/questions/7372038/is-there-any-way-to-change-input-type-date-format)

lamkeewei (approved):
lgtm! 

I added some styling to try to make it look closer to the design. Also added a default value for the date picker to resolve a warning complaining about it being missing.

PR #185, opened 2021-02-26
fix: lookup option by value instead of index for radio field

lamkeewei:
## Problem

Current Radio field implementation assumes that the `value` is equals to the index in the `options` array. However, this might not be true if the length of the `options` array changes. 

To reproduce the error:
1. Create a Radio field in the Question Tab
2. Delete the first option
3. Attempt to submit form in the Preview tab
4. An error with the message `Cannot read property 'label' of undefined` should be logged in the console

## Solution

**Bug Fixes**:

- Look up the corresponding option based on `value` when mapping user input to value for Radio fields.


prestonlimlianjie (approved):
lgtm

PR #206, opened 2021-03-04
feat: backend checker schema validation

lamkeewei:
## Problem

Currently, there is no check for whether a checker JSON is valid before saving it in the backend. This can lead to checkers that are invalid being saved.

## Solution

**Features**:
- Added `CheckerSchema` using `Joi` and run validation for `POST` and `PUT` endpoint
- Prevent deletion of last option for `Radio` and `Checkbox` field

## Tests
- Run unit tests in `CheckerSchema.spec.ts`
- Try to save an invalid checker on the frontend. An error should be show in the bottom right toast.

## Deploy Notes

**New dependencies**:

- `joi` : schema description and data validator

**New dev dependencies**:

- `@types/joi` : type definitions for `joi`


src/server/checker/CheckerController.ts
```diff
@@ -65,6 +75,13 @@ export class CheckerController {
     } else {
       try {
         const checker = req.body
+        const { error } = CheckerSchema.keys({
```
LoneRifle: It might be helpful to DRY this with the equivalent in `post()`

LoneRifle (approved):
lgtm otherwise

PR #211, opened 2021-03-04
feat: content security policy

barnettx:
Add content security policy (CSP) using helmet middleware to reduce chances of XSS attacks: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP. Implementation based off GoGovSG's CSP.

**New dependencies**:
- Helmet


src/server/bootstrap/helmet.ts
```diff
@@ -0,0 +1,19 @@
...
+const cspOnlyReportViolations = config.get('cspOnlyReportViolations')
+const cspReportUri = config.get('cspReportUri')
+
+export default helmet({
+  contentSecurityPolicy: {
+    directives: {
+      defaultSrc: ["'self'"],
+      styleSrc: ["'self'", "'unsafe-inline'", 'https://fonts.googleapis.com/'],
+      fontSrc: ["'self'", 'https://fonts.gstatic.com/'],
+      imgSrc: ["'self'", 'data:'],
```
barnettx: `'data:'` from Go. Don't think we have inline base64 images

src/server/bootstrap/helmet.ts
```diff
@@ -0,0 +1,19 @@
...
+import config from '../config'
+
+const cspOnlyReportViolations = config.get('cspOnlyReportViolations')
+const cspReportUri = config.get('cspReportUri')
+
+export default helmet({
+  contentSecurityPolicy: {
+    directives: {
+      defaultSrc: ["'self'"],
+      styleSrc: ["'self'", "'unsafe-inline'", 'https://fonts.googleapis.com/'],
```
barnettx, line 13: need to investigate further whether we use inline styles

src/server/bootstrap/helmet.ts
LoneRifle: Indeed we don't, so you can remove this.

src/server/bootstrap/helmet.ts
LoneRifle: Mostly by React

src/server/bootstrap/helmet.ts
```diff
@@ -0,0 +1,21 @@
...
+const cspReportUri = config.get('cspReportUri')
+const sentryDns = config.get('sentryDns')
+
+export default helmet({
+  contentSecurityPolicy: {
+    directives: {
+      defaultSrc: ["'self'"],
+      styleSrc: ["'self'", "'unsafe-inline'", 'https://fonts.googleapis.com/'],
+      fontSrc: ["'self'", 'https://fonts.gstatic.com/'],
+      connectSrc: ["'self'", ...(sentryDns ? [parseDomain(sentryDns)] : [])],
```
barnettx, line 15: maybe disable the linting here so that sentryDns is on a new line for better readibility?

LoneRifle: This is fine

src/server/bootstrap/helmet.ts
```diff
@@ -0,0 +1,21 @@
...
+      ...(cspReportUri ? { reportUri: cspReportUri } : {}),
```
LoneRifle, line 16: You didn't put the report Uri in an array in the end?

LoneRifle (changes requested):
Lgtm otherwise, but can you also enable CSP for the handler at `src/serverless/static.ts`? 

Once merged, push to `staging`, then verify that the changes work as intended.

src/server/bootstrap/helmet.ts
barnettx: In Sentry's example for how to set up report-uri, they don't include report-uri in the connect-src directive
```
def middleware(request, response):
    response['Content-Security-Policy-Report-Only'] = \
        "default-src 'self'; " \
        "report-uri https://o545501.ingest.sentry.io/api/5667596/security/?sentry_key=5049a239ada645f5900b090068b008bf"
    return response
```

I also tested it with a personal sentry i set up. The CSP reports go through.
image

barnettx: In Sentry's example for how to set up report-uri, they don't include report-uri in the connect-src directive
```
def middleware(request, response):
    response['Content-Security-Policy-Report-Only'] = \
        "default-src 'self'; " \
        "report-uri https://o545501.ingest.sentry.io/api/5667596/security/?sentry_key=XXXXXXXXXXXXXXXXX"
    return response
```

I also tested it with a personal sentry i set up. The CSP reports go through.
image

LoneRifle (approved):

barnettx:
reminder that CSP only works on local development when using server to serve front-end. webpack dev server does not use the helmet CSP
```
npm run build
npm run dev-server
```
then view access the app at the server port (8080)

PR #222, opened 2021-03-09
fix(auth): change user localStorage to User | null

LoneRifle:
## Problem

`useLocalStorage()` uses `JSON.parse()` to retrieve previously
stringified values from localStorage. This is problematic as
`JSON.stringify()` serializes `undefined` to the string "undefined",
which it is then unable to parse, throwing a spurious warning

## Solution

To mitigate this, change the type of `[user, setUser]` to be
`User | null`, so that we are forced to avoid `undefined`. This also
better lines up with what is returned by `/auth/whoami` in the backend.


lamkeewei (approved):
lgtm!

PR #223, opened 2021-03-09
fix: 'strech' typo in checker result align prop

barnettx:
- removes the chakra generated css for 'strech'
- no noticeable changes since the default for flex align-items is stretch


lamkeewei (approved):
lgtm! thanks for picking this up!

PR #233, opened 2021-03-11
fix: set checker id as random uuid string

### LINKED ISSUE #230: Set primary key of `checkers` table to be uuid instead of checker name
### Overview, Problem Description, Summary, or Abstract
Currently, the `checkers` table has a primary key of `id`. However, this `id` represents the name of the checker, as such, an admin user will not be able to rename their checker if they wanted to.

### Solution
Modify the primary key `id` to represent a random UUID and create a new `name` field to represent the checker name. Additionally, set the URL of the checker to use the new `id` instead of the `name` field so that we prevent broken links from happening whenever the checker name is modified.

prestonlimlianjie:
This PR sets the checker `id` as a random uuid string during checker creation instead of using the checker `title` as the `id`.

This allows us build features to allow admin users to rename their checkers in the future.

Fixes #230


package.json
```diff
@@ -89,6 +89,7 @@
     "@chakra-ui/react": "^1.3.3",
     "@emotion/react": "^11.1.4",
     "@emotion/styled": "^11.1.5",
+    "@types/uuid": "^8.3.0",
```
LoneRifle: this is a dev dependency. Given also that you are explicitly referencing uuid in your codebase, you probably want to `npm install uuid` as well.

LoneRifle (changes requested):

package.json
prestonlimlianjie: addressed in 270bbc6 and d1cc058

LoneRifle (approved):

PR #246, opened 2021-03-18
build(webpack): use asset modules, drop file-loader

LoneRifle:
Use the latest and greatest conventions for webpack 5


barnettx (approved):
lgtm

PR #247, opened 2021-03-19
feat: google analytics

barnettx:
## Problem

Currently, checkfirst does not have user analytics.

## Solution

I've added Google Analytics.

**Features**:

* Analytics tracks both visits to normal checkfirst and embedded checkfirst eg. checkfirst as an `iframe` in isomer sites
    * To allow iframe tracking, GA cookies are set to `SameSite=None; Secure`
    * Browsers have recently tightened cookie rules by ensuring the top-level domain (the domain in your browser navbar) is the same as the domain of the request before sending cookies in that request. In our case, the default `SameSite=Lax` for GA cookies meant that GA cookies were not being sent as the top-level domain eg. https://ogp-checkfirst-test-staging.netlify.app/ differed from the requests to http://www.google-analytics.com. `SameSite=None` tells the browser to ignore this same-domain check and attach the cookies on the GA request.
    * The default `SameSite=Lax` when creating cookies also meant that GA couldn't access cookie storage for its userIDs, leading to `analytics_debug.js:25 Cookie write failed.
analytics_debug.js:25 Storage not available. Aborting hit`
    * As far as I can tell, setting `SameSite=None` has minimal security implications
    * Because `SameSite=None` can only be set when the cookie is also `Secure` and the site is secure, this means that local testing will only work if webpack dev server has `https` enabled
    * References about SameSite cookies: https://web.dev/samesite-cookies-explained/ https://medium.com/trabe/cookies-and-iframes-f7cca58b3b9e
    * `react-ga` pointing out this issue: https://github.com/react-ga/react-ga/issues/457
* Injects the GA tracking ID during webpack build
* Added a basic user event of Submit when checker is submitted
* Implementation inspired by the Postman's (https://github.com/opengovsg/postmangovsg/pull/285) and Go's (https://github.com/opengovsg/GoGovSG/pull/754/files)

**Extensions**:
* Create plan of events to capture
* Consider upgrading to GA4. Didn't do so as it was released last year and doesn't seem to be mature

- `GA_TRACKING_ID` : Google Analytics Tracking ID that will be injected into the frontend during webpack build

**New dependencies**:

- `react-ga` : react google analytics library


src/client/contexts/GoogleAnalyticsContext.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export const GoogleAnalyticsProvider: FC = ({ children }) => {
+  const [isLoaded, setLoaded] = useState(false)
+
+  const location = useLocation()
+
+  useEffect(() => {
+    try {
+      initializeGA()
+    } catch (err) {
```
barnettx: Not sure if a try-catch is needed here. Copied this from: https://github.com/opengovsg/postmangovsg/pull/285/files#diff-9893cc2cc2b8bcca8c9660d02fc3422acf19cc738bb01fba2a1afcfc69c9f970L30-L32

src/client/services/GoogleAnalyticsService.tsx
```diff
@@ -0,0 +1,64 @@
+import ReactGA from 'react-ga'
+
+const GA_TRACKING_ID = process.env.GA_TRACKING_ID || ''
+
+const GA_USER_EVENTS = {
```
barnettx, line 5: maybe use a typescript enum?

src/client/contexts/GoogleAnalyticsContext.tsx
```diff
@@ -0,0 +1,78 @@
...
+      // TODO write error
```
LoneRifle: This is currently failing silently and swallowing the exception. If this is intentional, comment accordingly, otherwise, at the very least, log the exception, or let it pass through.

src/client/services/GoogleAnalyticsService.tsx
LoneRifle: Perhaps. On the other hand, now might be a good time to think through exactly how the GA events should be designed.

We have 4 groupings to think about - category, action, label, value. What can we map onto those?

Currently all GA events are categorised as `User`, with the only action available being `submit`. Could we perhaps incorporate more information from the submission into the event, perhaps the values of the outputs, and so on?

@lisatjide @lamkeewei - maybe something worth discussing here.

LoneRifle (changes requested):

LoneRifle (approved):
lgtm

serverless.yml
```diff
@@ -43,6 +43,7 @@ functions:
       SESSION_SECRET: ${env:SESSION_SECRET}
       OTP_SECRET: ${env:OTP_SECRET}
       OTP_EXPIRY: ${env:OTP_EXPIRY}
+      GA_TRACKING_ID: ${env:GA_TRACKING_ID}
```
lamkeewei: Is `GA_TRACKING_ID` required here?

src/client/services/GoogleAnalyticsService.tsx
```diff
@@ -0,0 +1,64 @@
...
+  ReactGA.initialize(GA_TRACKING_ID, {
+    debug: false, // Set to true only on development
+    useExistingGa: true,
+  })
+  ReactGA.ga('create', GA_TRACKING_ID, 'auto', {
+    cookieFlags: 'SameSite=None; Secure',
+  })
+}
+
+const setGAUserId = (userId: number | null): void => {
```
lamkeewei, line 19: I think this needs to be called to set the GA user ID to the logged in user ID after `whoami` or `verifyOtp` in the `AuthProvider`.

src/client/App.tsx
```diff
@@ -23,19 +27,21 @@ const App: FC = () => {
...
-              <PrivateRoute path="/dashboard" component={Dashboard} />
-              <PrivateRoute path="/builder/:id">
-                {/* TODO: Rename to BuilderProvider */}
-                <CheckerProvider>
-                  <FormBuilder />
-                </CheckerProvider>
-              </PrivateRoute>
-              <Redirect to="/" />
-            </Switch>
+            <GoogleAnalyticsProvider>
```
lamkeewei: Might want to move this up outside of `Router`. I remember having had issues if the Router's child is not a Route component. But can't confirm.

src/client/contexts/GoogleAnalyticsContext.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export const GoogleAnalyticsProvider: FC = ({ children }) => {
+  const [isLoaded, setLoaded] = useState(false)
+
+  const location = useLocation()
+
+  useEffect(() => {
+    initializeGA()
+    setLoaded(true)
```
lamkeewei: I think `isLoaded` is not required. [Hooks are evaluated in order](https://reactjs.org/docs/hooks-rules.html#explanation). So `initializeGA` should have been called before `sendPageView` in the next hook.

src/client/App.tsx
LoneRifle: Yup, that's indeed a problem

barnettx: Having GAProvider inside the Router doesn't seem to cause any issue. For postman it's the same https://github.com/opengovsg/postmangovsg/blob/develop/frontend/src/App.tsx#L31. Also I need to have the GAProvider inside the Router to use the `useLocation` hook that updates GA of a route change.

lamkeewei: Should be fine. I think the issue was when you have something that is not a `Route` in `Switch`.

lamkeewei (approved):
lgtm!

PR #251, opened 2021-03-19
fix: generate new uuid when duplicating checker

### LINKED ISSUE #245: Unable to duplicate checker
**Describe the bug**
Unable to duplicate checker

**To Reproduce**
1. Log in and click on plus icon at the bottom right on each card representing the checker.
Screenshot 2021-03-18 at 5 25 52 PM
 
2. Click 'Create' button
Screenshot 2021-03-18 at 2 21 26 PM

**Expected behavior**
There should not be error and modal for creating new checker should appear.

lamkeewei:
## Problem

Closes #245 

## Solution

_How did you solve the problem?_

**Improvements**:
- Allow user to modify checker title and description when duplicating checker

**Bug Fixes**:
- Generate a new `uuid` when duplicating checker

## Before & After Screenshots
image

## Tests
- Create a checker with title and description 
- Click duplicate and ensure that the form is pre-populated with values for title and description based on the original checker
- Confirm duplication and checker should be successfully created

### DIFF
--- a/src/client/components/dashboard/CreateNewModal.tsx
+++ b/src/client/components/dashboard/CreateNewModal.tsx
@@ -37,7 +37,6 @@ export const CreateNewModal: FC<CreateNewModalProps> = ({
   checker,
 }) => {
   const initial = {
-    id: uuidv4(), // Set id to be a random uuid string
     title: '',
     description: '',
     fields: [],
@@ -88,7 +87,7 @@ export const CreateNewModal: FC<CreateNewModalProps> = ({
     createChecker.mutate({
       ...(checker || initial),
       ...data,
-      id: uuidv4(),
+      id: uuidv4(), // Set id to be a random uuid string
     })
   }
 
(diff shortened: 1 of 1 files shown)


src/client/components/dashboard/CreateNewModal.tsx
```diff
@@ -77,7 +86,7 @@ export const CreateNewModal: FC<CreateNewModalProps> = ({
     description: string
   }) => {
     createChecker.mutate({
-      ...(checker || initial),
+      ...(checker ? { ...checker, id: uuidv4() } : initial),
```
LoneRifle: Actually why not...
```suggestion
      ...(checker || initial),
      id: uuidv4(),
```

Then eliminate the `id` field in `initial`

lamkeewei: Makes sense! Let me update that.

LoneRifle (approved):
you can also delete line 40 of this file, since the id will be injected only at time of creation.

lgtm otherwise

PR #272, opened 2021-03-23
fix: use ascii quotation marks

### LINKED ISSUE #241: Different quotation marks in generated iframe tag
**Describe the bug**
The iframe tag generated in checkfirst...
image
Uses a different quotation mark which doesn't register the css width correctly
image
image

**Expected behavior**
Use correct quotation marks

barnettx:
Closes #241 
- non-ascii quotation marks cause css in sample iframe tag in the embed modal to fail
- removed other instances of non-ascii quotes

Possible extension: add ESLint plugin: https://github.com/personaljs/eslint-plugin-ascii though i think this problem isn't significant enough to warrant the extra dependency

### DIFF
--- a/src/client/components/builder/Navbar.tsx
+++ b/src/client/components/builder/Navbar.tsx
@@ -208,7 +208,7 @@ export const Navbar: FC = () => {
                 name="iframe"
                 value={`<iframe src="${linkTo(
                   checker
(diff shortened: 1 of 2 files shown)


lamkeewei (approved):
lgtm!

PR #276, opened 2021-03-23
feat(builder): create DropdownField

LoneRifle:
## Problem

RadioField tends to be unwieldy for situations where more than 5 options are needed

## Solution

Create DropdownField, using textarea as input

- copy RadioField and add relevant entries across the builder and public views
- ditch the radio buttons borrowed from RadioField in favour of chakra's Select control
- use a textarea to accept DropdownField options

## Screenshots
image
image
image

### DIFF
--- a/src/client/components/builder/QuestionsTab.tsx
+++ b/src/client/components/builder/QuestionsTab.tsx
@@ -81,7 +81,7 @@ export const QuestionsTab: FC = () => {
   useEffect(() => {
     let highestIndex = 0
     config.fields.forEach((field) => {
-      const fieldIndex = parseInt(field.id.slice(1))
+      const fieldIndex = parseInt((field.id.match(/\d+/) || [])[0])
       highestIndex = Math.max(highestIndex, fieldIndex)
     })
     setNextUniqueId(highestIndex + 1)
--- a/src/client/components/builder/QuestionsTab.tsx
+++ b/src/client/components/builder/QuestionsTab.tsx
@@ -8,6 +8,7 @@ import {
   BiDownArrowAlt,
   BiCalendar,
 } from 'react-icons/bi'
+import { IoIosArrowDropdown } from 'react-icons/io'
 import { VStack } from '@chakra-ui/react'
 
 import * as checker from '../../../types/checker'
@@ -18,6 +19,7 @@ import {
   CheckboxField,
   TitleField,
   DateField,
+  DropdownField,
 } from '../builder/questions'
 
 import { useCheckerContext } from '../../contexts'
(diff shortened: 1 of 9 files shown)


src/client/components/builder/questions/DropdownField.tsx
```diff
@@ -0,0 +1,106 @@
...
+    <VStack align="stretch" w="100%" spacing={6}>
+      <VStack align="stretch">
+        <HStack>
+          <BiListUl fontSize="20px" />
+          <Text>{title}</Text>
+        </HStack>
+        {description && <Text color="#718096">{description}</Text>}
+      </VStack>
+      <Select>
+        {options.map(({ value, label }, i) => (
```
lamkeewei, line 95: Just wondering, do you think we should enforce some sort of sort order on the select options? A really long list (e.g. countries) would benefit from it. 

Along the same vein, should we try to create a select component for this with search, similar to Forms? This can be done with `downshift` and `match-sorter` like in [ExpressionInput](https://github.com/opengovsg/checkfirst/blob/develop/src/client/components/builder/logic/ExpressionInput.tsx).

LoneRifle: My worries are as follows:
- there might be a reason why the user may wish to impose his/her own sort order. even for something like a list of countries, the user may decide to prioritise countries where we expect a lot of traffic (see [The Oatmeal](https://theoatmeal.com/comics/shopping_cart))
- I think we could! My suspicion is that Chakra deliberately used the native form options in the interest of accessibility. If there is a way to do it such that we have a more consistent visual element while still fulfilling accessibility goals, we could invest some time in it

src/client/components/builder/QuestionsTab.tsx
```diff
@@ -42,6 +44,14 @@ const generateDefaultRadioField = (id: number): checker.Field => ({
   options: [{ label: 'Option 1', value: 0 }],
 })
 
+const generateDefaultDropdownField = (id: number): checker.Field => ({
+  id: `DL${id}`,
```
lamkeewei, line 48: Currently when you duplicate, the ID will be `DX` instead of `DLX`. This is an issue because of [our hacky way](https://github.com/opengovsg/checkfirst/blob/122c8e4b97ca17826d7f5d3cfa0337ebf9f71694/src/client/components/builder/BuilderField.tsx#L145) of doing duplicates which assume that the prefix for a field type is just 1 character long.

src/client/components/builder/questions/DropdownField.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return (
+    <VStack align="stretch" w="100%" spacing={6}>
+      <VStack align="stretch">
+        <HStack>
+          <BiListUl fontSize="20px" />
+          <Text>{title}</Text>
+        </HStack>
+        {description && <Text color="#718096">{description}</Text>}
+      </VStack>
+      <Select>
```
lamkeewei:
```suggestion
      <Select isDisabled sx={styles.dummyInput}>
```

Where `styles = useStyles()`.  Minor nit to standardise the styles for preview component and also to disable the input to prevent triggering the dropdown when clicking on the card to enter into the input state.

jumpy-select

lamkeewei (commented):
Just minor issues with duplication and preview component for dropdown. I'm ok with keeping to the Chakra select for now.

src/client/components/builder/QuestionsTab.tsx
LoneRifle: Fixed that here:
https://github.com/opengovsg/checkfirst/pull/276/files#diff-15a8ab6b16cf6cdcbfc83b28f231671796c54b4641f02b0d4bd96be8f53801a4R84

lamkeewei (approved):
lgtm!

PR #278, opened 2021-03-24
feat: add backend sentry

barnettx:
- use sentry express integration
- set user from session user

- `BACKEND_SENTRY_DSN`
- `CSP_REPORT_URI`: wasn't previously linked up in `serverless.yml` and `ci.yml` for the CSP PR

**New dependencies**:

- `@sentry/node`
- `@sentry/tracing`

### Testing

Add a route that throws an error: https://docs.sentry.io/platforms/node/guides/express/. Then ping that endpoint by navigating to it on browser eg. `localhost:3000` or using postman.
```
app.get("/debug-sentry", function mainHandler(req, res) {
  throw new Error("My first Sentry error!");
});
```

NOTE: an error thrown in a async function won't be caught by sentry error handler

TODO: 
- fix `sentrySessionMiddleware` to add user info when error occurs
- investigate further as to why async functions are not caught by sentry

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,6 @@ jobs:
             echo SESSION_SECRET=${{ secrets.SESSION_SECRET_STAGING }} >> $GITHUB_ENV;
             echo OTP_SECRET=${{ secrets.OTP_SECRET_STAGING }} >> $GITHUB_ENV;
             echo GA_TRACKING_ID=${{ secrets.GA_TRACKING_ID_STAGING }} >> $GITHUB_ENV;
-            echo SENTRY_ENVIRONMENT=${{ secrets.SENTRY_ENVIRONMENT_STAGING }} >> $GITHUB_ENV;
             echo APP_HOST=staging.checkfirst.gov.sg >> $GITHUB_ENV;
           elif [[ $GITHUB_REF == $PRODUCTION_BRANCH ]]; then
             echo LAMBDA_NODE_ENV=production >> $GITHUB_ENV;
@@ -111,7 +110,6 @@ jobs:
             echo SESSION_SECRET=${{ secrets.SESSION_SECRET_PRODUCTION }} >> $GITHUB_ENV;
             echo OTP_SECRET=${{ secrets.OTP_SECRET_PRODUCTION }} >> $GITHUB_ENV;
             echo GA_TRACKING_ID=${{ secrets.GA_TRACKING_ID_PRODUCTION }} >> $GITHUB_ENV;
-            echo SENTRY_ENVIRONMENT=${{ secrets.SENTRY_ENVIRONMENT_PRODUCTION }} >> $GITHUB_ENV;
             echo APP_HOST=checkfirst.gov.sg >> $GITHUB_ENV
           fi
           echo SENTRY_AUTH_TOKEN=${{ secrets.SENTRY_AUTH_TOKEN }} >> $GITHUB_ENV;
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,3 +137,4 @@ jobs:
           APP_HOST: ${{ env.APP_HOST }}
           FRONTEND_SENTRY_DSN: ${{ secrets.FRONTEND_SENTRY_DSN }}
           BACKEND_SENTRY_DSN: ${{ secrets.BACKEND_SENTRY_DSN }}
+          CSP_REPORT_URI: ${{ secrets.CSP_REPORT_URI }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,6 +116,8 @@ jobs:
         env: 
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
           FRONTEND_SENTRY_DSN: ${{ secrets.FRONTEND_SENTRY_DSN }}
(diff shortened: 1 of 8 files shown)


LoneRifle (changes requested):
Lgtm otherwise - merge when I've stashed the necessary DSNs

src/server/bootstrap/index.ts
```diff
@@ -72,20 +74,55 @@ export async function bootstrap(): Promise<Express> {
...
+        email: email,
+      })
+    }
+    next()
+  }
+
   const app = express()
 
+  Sentry.init({
+    dsn: process.env.BACKEND_SENTRY_DSN,
```
LoneRifle: Could you feed this via convict please?

LoneRifle (changes requested):
Lgtm otherwise, will merge once Sentry set up

.github/workflows/ci.yml
```diff
@@ -116,6 +116,7 @@ jobs:
           fi
           echo SENTRY_AUTH_TOKEN=${{ secrets.SENTRY_AUTH_TOKEN }} >> $GITHUB_ENV;
           echo FRONTEND_SENTRY_DSN=${{ secrets.FRONTEND_SENTRY_DSN }} >> $GITHUB_ENV;
+          echo BACKEND_SENTRY_DSN=${{ secrets.BACKEND_SENTRY_DSN }} >> $GITHUB_ENV;
           echo SENTRY_ORG=${{ secrets.SENTRY_ORG }} >> $GITHUB_ENV;
           echo SENTRY_PROJECT=${{ secrets.SENTRY_PROJECT }} >> $GITHUB_ENV;
           echo SENTRY_URL=${{ secrets.SENTRY_URL }} >> $GITHUB_ENV;
```
LoneRifle: Given that these are no longer needed, can we have them removed?

serverless.yml
```diff
@@ -44,6 +44,8 @@ functions:
       OTP_SECRET: ${env:OTP_SECRET}
       OTP_EXPIRY: ${env:OTP_EXPIRY}
       FRONTEND_SENTRY_DSN: ${env:FRONTEND_SENTRY_DSN}
+      BACKEND_SENTRY_DSN: ${env:BACKEND_SENTRY_DSN}
+      SENTRY_ENVIRONMENT: ${env:SENTRY_ENVIRONMENT}
```
LoneRifle: I think instead of doing this, consider recycling `NODE_ENV` as defined above by removing this line like thus, then injecting ...
```suggestion

```

src/server/bootstrap/index.ts
```diff
@@ -72,20 +74,55 @@ export async function bootstrap(): Promise<Express> {
...
+    // automatically picks up environment from SENTRY_ENVIRONMENT
+    integrations: [
+      // enable HTTP calls tracing
+      new Sentry.Integrations.Http({ tracing: true }),
+      // enable Express.js middleware tracing
+      new Tracing.Integrations.Express({ app }),
+    ],
+    tracesSampleRate: 1.0,
```
LoneRifle, line 102: ... by injecting it as `environment:` here, ie 
```suggestion
    tracesSampleRate: 1.0,
    environment: config.get('nodeEnv'),
```

LoneRifle (changes requested):
Sorry, found something else while trying to configure the environment

.github/workflows/ci.yml
barnettx: these are needed by the sentry webpack plugin. will rename them to clarify that they are for the frontend sentry project.

LoneRifle: If I'm not mistaken, `SENTRY_URL` defaults to https://sentry.io?

barnettx: just remembered why i didn't rename them earlier. because sentry automatically looks for the env vars and adds to them to the config. i'll add them to config explicitly so it's clearer

barnettx: yea looking at the docs again, it seems like most of these envs are optional: https://github.com/getsentry/sentry-webpack-plugin. we just need SENTRY_DSN, SENTRY_AUTH_TOKEN, and environment set through NODE_ENV

LoneRifle: Specify the env as `LAMBDA_NODE_ENV`. Never figured out why webpack couldn't build when `NODE_ENV` was set to `production`

barnettx: turns out `SENTRY_ORG` and `SENTRY_PROJECT` are required
image

barnettx:
User info will be captured using this test error route. Had to include `sessionMiddleware`
```
  const apiMiddleware = [
    sessionMiddleware,
    bodyParser.json(),
    sentrySessionMiddleware,
  ]
app.get("/debug-sentry", apiMiddleware, function mainHandler(req, res) {
  throw new Error("sentry error with user info");
});
```

barnettx:
> User info will be captured using this test error route. Had to include `sessionMiddleware`
> 
> ```
>   const apiMiddleware = [
>     sessionMiddleware,
>     bodyParser.json(),
>     sentrySessionMiddleware,
>   ]
> app.get("/debug-sentry", apiMiddleware, function mainHandler(req, res) {
>   throw new Error("sentry error with user info");
> });
> ```

Sentry will show:
image

barnettx:
Latest branch has been tested for both sentry frontend and backend using personal sentry account. Also tested CSP reporting.

LoneRifle (approved):

PR #279, opened 2021-03-25
feat: create checker from template

lamkeewei:
## Problem

Creating a template from a blank state might be intimidating for users. We should allow them to get started with some common templates that they can modify from.

## Solution

**Features**:
- Added `templates` table to store templates
- Added `/template` endpoints to allow fetching of templates
- Implemented frontend flow for creation of checker from template

**Improvements**:
- Refactored creation modal so that it can be reused for creation from template or blank checker.

## Before & After Screenshots

image

image

## Tests
**Create modal**
- Clicking on "Create from scratch" should show the creation form
- Selecting one of the template should redirect to the preview

**Template preview**
- Preview checker should be fully functional (able to key in inputs and submit results)
- "Back to templates" should redirect back to the template selector modal
- Close button (X) on the extreme right should redirect back to the dashboard
- "Use template" should direct to the creation form with title and description pre-populated based on the template.

## Deploy Notes

**Database migrations**:
- Run `npx sequelize db:migrate` to create new `templates` table
- Run `npx sequelize db:seed:all` to seed the database with the initial set of templates.


src/server/database/models/Template.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  @Column({
+    type: DataType.TEXT,
+    allowNull: true,
+  })
+  description!: string
+
+  @Column({
+    type: DataType.JSON,
+    allowNull: true,
```
prestonlimlianjie: Should the `fields`, `constants`, `operations`, and `displays` columns be set to `allowNull: false` to match the migrations? I realized that I made this discrepancy in `Checkers` as well so I modified the model file in the add-preview-publish PR https://github.com/opengovsg/checkfirst/pull/325/files#diff-23e6de5c559cc75fc240a917fe0c9832b8beb0dcd5d03d40dfc5495a4d853d12

src/server/bootstrap/index.ts
```diff
@@ -53,6 +54,10 @@ export async function bootstrap(): Promise<Express> {
     }),
   })
 
+  const template = new TemplateController({
+    service: new TemplateService(),
```
prestonlimlianjie, line 58: ~Note to self: do we need to include `logger`?~

This is a comment, not a fix:

We don’t really use logger consistently in our backend code now. This is probably a separate discussion/issue - we should think about what exactly we want to log/or if we should just log every single service method call

Created issue https://github.com/opengovsg/checkfirst/issues/339

prestonlimlianjie (commented):
lgtm - had a comment about the `allowNull` condition in the `Templates` model

prestonlimlianjie (approved):
lgtm

PR #285, opened 2021-03-27
refactor: use sequelize-typescript

prestonlimlianjie:
## Introduction
This PR is a proposal to refactor our codebase to use `sequelize-typescript`. This is the second of 3 PRs; please review https://github.com/opengovsg/checkfirst/pull/283 before this PR.

We discuss 3 major engineering design decisions below.

### We want to modify our codebase to use `sequelize-typescript` instead of `sequelize` because the former allows us to express tables/models in a declarative manner.

This is great because there is no more second-guessing what sequelize is doing under the hood w.r.t. associations.

**Associations are declared within the model file so we don’t need to jump to other files to find it (the `BelongsToMany` declaration is made explicit)**
```
// ./src/server/database/models/user.ts

@Table({ tableName: 'users', timestamps: true })
export class User extends Model<User> {
  @Column({
    primaryKey: true,
    type: DataType.INTEGER,
    autoIncrement: true,
  })
  id!: number

  @Column({
    type: DataType.TEXT,
    unique: true,
    allowNull: false,
    validate: {
      isEmail: true,
      isLowercase: true,
      is: emailValidator.makeRe(),
    },
    set(this: Settable, email: string) {
      // must save email as lowercase
      this.setDataValue('email', email.trim().toLowerCase())
    },
  })
  email!: string

  @BelongsToMany(() => Checker, () => UserToChecker)
  checkers!: Checker[]
}

```

**Foreign keys/join tables are explicitly defined as their own model file as well (instead of being implicitly defined by the `.belongsToMany({ through: <join_table> })` method)**

```
// ./src/server/database/models/UserToChecker.ts

@Table({ tableName: 'usersToCheckers', timestamps: true })
export class UserToChecker extends Model<UserToChecker> {
  @ForeignKey(() => User)
  @Column
  userId!: number

  @ForeignKey(() => Checker)
  @Column
  checkerId!: string
}
```

### We want to move the code to initialize models for the sequelize instance into the `bootstrap/sequelize.ts` file

We do this because we view the step of adding models to the sequelize instance as a logical extension of the other configurations (e.g. setting timezone, DB url, dialect). Moreover, `sequelize-typescript` allows us to do it quite elegantly using a [syntactic-sugar method](https://github.com/RobinBuschmann/sequelize-typescript#configuration), `.addModels()`. So now, we can very easily see our full sequelize config (what database, what dialect, what models are added) in just a few lines of code:

```
let sequelize = databaseUrl
  ? new Sequelize(databaseUrl, {
      timezone: '+08:00',
      logging: logger.info.bind(logger),
    })
  : new Sequelize({
      dialect: 'sqlite',
      ...(sqlitePath ? { storage: sqlitePath } : {}),
    })

sequelize.addModels([User, Checker, UserToChecker])
export default sequelize
```

### We want to modify the codebase to look to the model class as the main source of types
This idea is a little abstract, so let me explain it with specific references to our code. Take for example, the User type. At present, our code only has 1 representation of the `User` type - and this type is used in both our application (backend services and frontend) layer as well as our database model layer.

However, we need at least 2 representations of the `User` type because the application and model layers really interact with 2 slightly different versions of `User`.

The application layer sees this:
```
{
  id: number
  email: string
}
```

Whereas the model layer sees this:
```
{  
  id: number
  email: string
  checkers: Checker[]
}
```

Our current type for User is:
```
export interface User {
  id: number
  email: string
}
```

As you can see, the `User` type is really just the application layer’s definition of `User`. And as a result, our current codebase uses the application layer as the source of types.

This causes several problems especially now that we are using `sequelize-typescript` to explicitly declare all columns in our model class. Specifically, we are not able to shoehorn the existing `User` type into the model type.

Instead, this PR proposes to use the model’s type definition as the source, and create a new type definition for the application layer using the `Pick` utility type. Here is an example for `User`. We now have 2 types for the model and application representation of `User`, and the source of the types comes from the model class:
```
export class ModelUser extends Model<User> {
  id: number
  email: string
  checkers: Checker[]
}

type ApplicationUser = Pick<ModelUser, "id", "email">
```

## Note: there is some jankiness in the `Model.create()` code

I had to do this strange type assertion chain [here](https://github.com/opengovsg/checkfirst/compare/refactor/use-sequelize-typescript?expand=1#diff-67163cff3b0980c0ba50b4cebb47ed61a1e1420123aab746629afa7dd95b61bfR41):
`const checkerInstance = await this.CheckerModel.create((checker as any) as CheckerModel, options)`

The original code was:
`const checkerInstance = await this.CheckerModel.create(checker, options)`

However, I was getting this error message:
```
src/server/checker/CheckerService.ts:46:64 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(this: ModelStatic<Checker>, values?: Checker | undefined, options?: CreateOptions<Checker> | undefined): Promise<Checker>', gave the following error.
    Argument of type 'import("/Users/prestonlim/Documents/OGP/checkfirst/src/types/checker").Checker' is not assignable to parameter of type 'import("/Users/prestonlim/Documents/OGP/checkfirst/src/server/database/models/Checker").Checker'.
      Type 'Checker' is missing the following properties from type 'Checker': $add, $set, $get, $count, and 32 more.
  Overload 2 of 2, '(this: ModelStatic<Checker>, values: Checker, options: CreateOptions<Checker> & { returning: false; }): Promise<void>', gave the following error.
    Argument of type 'import("/Users/prestonlim/Documents/OGP/checkfirst/src/types/checker").Checker' is not assignable to parameter of type 'import("/Users/prestonlim/Documents/OGP/checkfirst/src/server/database/models/Checker").Checker'.
```

This part of the error message is particularly confusing, and there’s still an [unresolved open issue on the repo](https://github.com/RobinBuschmann/sequelize-typescript/issues/936) 
```
`Type 'Checker' is missing the following properties from type 'Checker': $add, $set, $get, $count, and 32 more.
```

I followed the example on the thread to do a type assert to any and then back to the CheckerModel - but I acknowledge that it is janky and we should fix in the future.

## How to test
The goal of the test is to figure out if changes to the model in this PR are compatible with the existing tables in the database:
- Ensure that your local DB schemas are in an old state
- Run `npm run dev` and test the application


tsconfig.json
```diff
@@ -5,7 +5,7 @@
 
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
-    "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
+    "target": "es6",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
```
LoneRifle, line 8: Leaving a note here for our future selves - this is okay because we have a `src/client/tsconfig.json` that directs `tsc` to target ES5 (which we will need for older browsers)

.sequelizerc
```diff
@@ -0,0 +1,8 @@
+const path  = require('path')
+
+module.exports = {
+    "config": path.resolve('./build/server/database/config', 'config.js'),
+    "models-path": path.resolve('./build/server/database/models'),
+    "seeders-path": path.resolve('./build/server/database/seeders'),
+    "migrations-path": path.resolve('./src/server/database/migrations')
+};
```
LoneRifle, line 8: This isn't linted, but I'm going to trust that Sequelize itself is sensitive to its own formatting

LoneRifle (approved):

LoneRifle:
lgtm, merge when ready

PR #289, opened 2021-03-29
fix(builder): do not fetch checker on window refocus

LoneRifle:
## Problem
By default, react-query fetches the last saved state of the checker
when the user refocuses the window (eg, by clicking on another browser
tab then returning to CheckFirst). This decimates any unsaved changes
that the user might have had.

## Solution
Fix this by explicitly specifying no fetching on window refocus


lamkeewei (approved):
lgtm!

PR #295, opened 2021-03-29
fix(builder): ensure deep copy of field, remove code repetition

### LINKED ISSUE #287: Editing the option of a duplicated question will also edit the option from the original question
**Describe the bug**
Duplicate a question on the form. Start editing one of the option of the duplicated question. 
You will notice that both will be updated.

Screenshot 2021-03-29 at 11 15 55 AM

**Expected behavior**
Editing the duplicated option should not also edit the original option.

LoneRifle:
## Problem

Closes #287 

## Solution

- Make deep copy of field through `JSON.stringify()` and `JSON.parse()`
- DRY `handleDuplicate()` out by limiting repeated code to the creation
  of `payload`. Duplicate blocks are still needed so that we match the
  correct ConfigArrayEnum to the correct type of `element`


prestonlimlianjie (approved):
lgtm - could we leave a short comment in code on the `...JSON.parse(JSON.stringify(data))` line?

PR #297, opened 2021-03-29
feat: sentry error boundary with fallback component

barnettx:
Sentry error boundary to catch any frontend errors and displays fallback instead of a possibly mangled frontend due to error containing react state. https://docs.sentry.io/platforms/javascript/guides/react/components/errorboundary/

Fallback component contains a button that allows the app to return to the state before the error occurred.

**Fallback screen**
Screenshot 2021-03-29 at 4 14 47 PM

## Testing ##

Throwing an error using a button as suggested in https://github.com/isomerpages/isomercms-frontend/pull/262 won't trigger the error boundary, though the error will be sent
```
<button onClick={() => { throw new Error('hello')}}>Err</button>
```

Instead use a error producing component from the react error boundary docs: https://reactjs.org/docs/error-boundaries.html

```
class BuggyCounter extends React.Component<{}, {counter: number}> {
  constructor(props: {}) {
    super(props);
    this.state = { counter: 0 };
    this.handleClick = this.handleClick.bind(this);
  }
  
  handleClick() {
    this.setState(({counter}) => ({
      counter: counter + 1
    }));
  }
  
  render() {
    if (this.state.counter === 1) {
      // Simulate a JS error
      throw new Error('I crashed!');
    }
    return <h1 onClick={this.handleClick}>{this.state.counter}</h1>;
  }
}
```

In page render, eg. Dashboard.tsx

```
...
<Flex direction="column" minH="100vh" bgColor="#F4F6F9">
      <Navbar />
      <Container maxW="960px" pt="80px" px={0}>
        <BuggyCounter />
        <VStack align="stretch" py={10} position="relative">
          {isLoading ? (
            <Center>
...
```
Clicking on the counter will then trigger the error and hence the error boundary.

Note: the `BuggyCounter` differs from the above button because the error is triggered during the `render` method. Error boundaries **DO NOT** catch errors in event handlers: https://reactjs.org/docs/error-boundaries.html#how-about-event-handlers


prestonlimlianjie (changes requested):
For readability and maintainability, let's define the `fallback` component as its own `Page` or `Component`

barnettx:
> Good PR description, especially w.r.t. how you perform testing. For readability and maintainability, let's define the `fallback` component as its own `Page` or `Component`

Good catch. Made the refactor.

prestonlimlianjie (approved):
lgtm

PR #298, opened 2021-03-29
chore: add experimentalDecorators config in client

prestonlimlianjie:
This PR sets the `experimentalDecorators` flag to `true` in the client's `tsconfig.json` to suppress the following compiler errors. 
Screenshot 2021-03-29 at 5 33 02 PM

Note: We attempted to use the `exclude` parameter: `exclude: ['../server']` to no avail.


lamkeewei (approved):
lgtm! though still strange exclude doesn't work. :/

PR #299, opened 2021-03-29
fix(builder): disable duplicate and delete on title card

### LINKED ISSUE #291: Page turns blank when the use click on the Bin on the checker title
**Describe the bug**
Page turns blank when user click on the Bin icon on the checker title. 

Screenshot 2021-03-29 at 3 08 14 PM

**Expected behavior**
There should not be a bin button on the title
or
The bin could represent the act of deleting the checker. If that's the case, a pop up to confirm should appear, and user will be brought back to the dashboard.

lamkeewei:
## Problem

Title cards should not have duplicate or delete options. All checkers should have one and only one title.

Closes #291 

## Solution

**Bug Fixes**:

- Skip rendering of duplicate and delete options if it is a title card

## Before & After Screenshots
image


prestonlimlianjie (approved):
lgtm

PR #302, opened 2021-03-30
docs(changelog): add auto-changelog, npm version hook

LoneRifle:
## Problem

We don't currently have a proper changelog to track our changes. As we start gaining users 
and the codebase stabilises, we will need to publish the changes we have made to the codebase in each release

## Solution
- Add auto-changelog and a hook to the `npm version` command
- Generate CHANGELOG.md, assuming `v1.0.0` to be what is currently in `production`

This PR is based on opengovsg/GoGovSG#891. See that PR for further information on using auto-changelog and npm version patch when incrementing the version number during the release cycle.


prestonlimlianjie (approved):
lgtm

PR #311, opened 2021-03-31
fix: missing db hostname and added CSP env var to static lambda

lamkeewei:
## Problem
- Database hostname was missing in the connection config
- CSP headers env var for static lambda were missing


prestonlimlianjie (approved):
lgtm

Issue #317, opened 2021-04-01
Set `cookieMaxAge` to 24 hours


prestonlimlianjie:
Requested by @lisatjide

PR #320, opened 2021-04-01
fix: set default cookieMaxAge to 24h

prestonlimlianjie:
Fixes https://github.com/opengovsg/checkfirst/issues/317


lamkeewei (approved):
lgtm!

PR #323, opened 2021-04-05
fix(radio): avoid selecting default value

LoneRifle:
## Problem

Checkers currently render radio button groups with the first value selected,
potentially leading the user

## Solution

To avoid leading the user of a checker, do not set the form radio
option to any default value


lamkeewei (approved):
lgtm!

Issue #300, opened 2021-03-29
Add a preview-publish flow


prestonlimlianjie:
## Is your feature request related to a problem? Please describe.
Currently, when the admin user modifies and saves a checker, the checker is immediately published. We would like to add a new flow that allows the admin user to 1) preview the checker and then 2) explicitly publish the checker when they are ready to do so.

## Describe the solution you'd like
Design done by Natalie: https://www.figma.com/file/qtoA63mFtCW3jGYhr8ev7y/Master?node-id=1417%3A15883

#### Database/Model layer
- Create a new `PublishedCheckers` table that contains the same basic fields as the `Checkers` table and is associated to a Checker
- There will be a 1-to-many relationship between Checkers and PublishedCheckers

#### API routes layer
```
// New endpoints
  api.get('/c/:id', checker.getPublished)
  api.post('/c/drafts/:id/publish', checker.publish)
  api.post('/c/drafts', checker.post)
  api.get('/c/drafts', checker.list)
  api.get('/c/drafts/:id', checker.get)
  api.put('/c/drafts/:id', checker.put)
  api.delete('/c/drafts/:id', checker.delete)
```

#### Services layer
- Create new `checker.getPublished()` method to return the latest version of the published checker
- Create new `checker.publish()` method to publish existing checker draft

#### Migrations
- Create the relevant migrations for the database

lisa-kwok:
appending the working doc for this thread[ here](https://docs.google.com/document/d/1rpqjfWTNcb_Eqv_a8mRl4hrKphIcItr7rzLabcYf3Ws/edit#heading=h.p0igmd5ylcux)

PR #325, opened 2021-04-05
feat: add preview publish feature

prestonlimlianjie:
Closes https://github.com/opengovsg/checkfirst/issues/300

### PublishedCheckers table
We model the `Checker`:`PublishedChecker` relationship as a **1:m** one. The `PublishedCheckers` table is treated as an append-only table where we create a new record whenever a user publishes a new checker version.

The public version of the checker is the most-recently-created record in the `PublishedCheckers` table.

### API design
Refer to #300 for new APIs

### API Service method for `CheckerService.publish`
We designed `CheckerService.publish` such that it does the following:
- updates the record in the `Checkers` table with the incoming checker
- creates a new record in the `PublishedCheckers` table with the incoming checker

We do so instead of taking the existing checker in the `Checkers` table and save that into the `PublishedCheckers` table; this is because there could be race conditions if there were multiple collaborators on the project. 

Consider the following sequence of events:
- user A saves their checker
- user B saves a new checker
- user A clicks on publish

If the publish method copied the existing checker in the `Checkers` table, we would end up publishing user B's version of the checker instead of user A's. This could be disastrous as user B might have just saved their checker - but it might not be a version that was ready to publish.

#### Refactor: EmbedModal
We created an `EmbedModal` component that is reused across the `Navbar` and `PreviewNavBar` components.

#### Refactor: Use `CheckerService` methods in the frontend
Clean up `src/client/pages/Dashboard.tsx` to use `CheckerService.list` instead of calling the raw API.

## Deploy/release notes
- Run database migration 
- Run SQL queries to publish the selected few checkers that are meant to be published

---

#### New button layout on dashboard navbar: preview, save draft, publish
Screenshot 2021-04-05 at 6 26 51 PM

#### Preview mode navbar
Screenshot 2021-04-05 at 6 26 53 PM

#### Modified alert message in embed modal
Screenshot 2021-04-05 at 6 27 02 PM


prestonlimlianjie:
I was unable to add a hyperlink to the toast for the "successfully published checker" message. Added an issue to be resolved in the near future https://github.com/opengovsg/checkfirst/issues/326.

src/server/checker/CheckerService.ts
```diff
@@ -68,19 +79,79 @@ export class CheckerService {
...
-        return resultWithoutUsers
+    return await this.findAndCheckAuth(id, user)
+  }
+
+  retrievePublished: (
+    id: string
+  ) => Promise<GetPublishedCheckerWithoutDraftCheckerDTO | undefined> = async (
+    id
+  ) => {
+    const result = await this.PublishedCheckerModel.findOne({
```
lamkeewei, line 90:
```suggestion
    const result = await this.PublishedCheckerModel.findOne({
        where: { checkerId: id },
        order: [['createdAt', 'DESC']],
     })
```

Would this work? This way we don't have to manually remove included `checker` later on.

src/server/checker/CheckerService.ts
```diff
@@ -68,19 +79,79 @@ export class CheckerService {
...
+        description: checker.description,
+        fields: checker.fields,
+        constants: checker.constants,
+        operations: checker.operations,
+        displays: checker.displays,
+        checkerId: id, // we want to make sure that we are publishing the right checker
+      }
+
+      await this.PublishedCheckerModel.create(createPublishedChecker, {
+        include: [this.CheckerModel],
```
lamkeewei:
```suggestion
```
From what I get from the [docs](https://sequelize.org/master/manual/creating-with-associations.html), I think this might not be needed since we're not creating the Checker instance also with this call.

src/server/checker/CheckerService.ts
```diff
@@ -68,19 +79,79 @@ export class CheckerService {
...
+  ) => Promise<Checker> = async (id, checker, user) => {
+    const transaction = await this.sequelize.transaction()
+    const transactionOptions = this.isSqliteFile ? {} : { transaction }
+
+    try {
+      const existingChecker = await this.findAndCheckAuth(id, user)
+      if (!existingChecker) throw new Error(`Checker ${id} not found`)
+
+      // Perform update in checkers and create in publishedCheckers with 1 transaction
+      await this.CheckerModel.update(checker, {
```
lamkeewei, line 120: Not critical, can consider encoding this constraint within the `PublishedChecker` model using Sequelize's `BeforeCreate` [hook](https://github.com/RobinBuschmann/sequelize-typescript#hooks).

lamkeewei (commented):
Pending tests and left some minor comments on `sequelize` usage. 

Also, might want to include a database migration to perform a one-time copy from `checkers` to `publishedCheckers`. Given that our checkers are currently published by default, this should be fine. 

When releasing this new feature, we should remember to run the migrations first before deploying the code change. Otherwise, there will be a slight downtime for all active checkers since they can't be found.

Otherwise, tested the following and works as expected:
- Preview should show last saved draft checker
- Public checker page should show last published checker
- Accessing a yet-to-be published checker should return an error

src/server/checker/CheckerService.ts
prestonlimlianjie: Making a commit with your suggestion because it is cleaner.

I also realized that we still needed to rename `checkerId` to `id` before returning the DB object. We can resolve this problem using the the nested array syntax in the `attributes` option .

Fixed in f9acf1f

```
    const result = await this.PublishedCheckerModel.findOne({
      attributes: [
        ['checkerId', 'id'], // rename checkerId as id 
        'title',
        'description',
        'fields',
        'constants',
        'operations',
        'displays',
      ],
      where: { checkerId: id },
      order: [['createdAt', 'DESC']],
    })

```

prestonlimlianjie:
> Pending tests and left some minor comments on `sequelize` usage.
> 
> Also, might want to include a database migration to perform a one-time copy from `checkers` to `publishedCheckers`. Given that our checkers are currently published by default, this should be fine.

- Tests added in 9862e34
- I had a chat with Lisa re: migrating all checkers to be published; she said that we should only publish the ICA SafeTravel checkers

src/server/checker/CheckerService.ts
prestonlimlianjie: Fixed in 0882214

src/server/checker/CheckerService.ts
prestonlimlianjie: Fixed in 0882214

src/server/checker/CheckerService.ts
prestonlimlianjie: Will explore this in a separate PR - thanks for the suggestion! Created issue https://github.com/opengovsg/checkfirst/issues/341.

lamkeewei (approved):
lgtm!

PR #360, opened 2021-04-09
fix: make active input bg white

### LINKED ISSUE #333: Misleading input field color
### Problem
Some of the input fields in logic blocks are filled with the grey background, which is also used in the question tabs as disabled fields. Issue raised by @NatMaeTan.

Screenshot 2021-04-06 at 2 53 17 PM

### Fix
We should not be using filled background fields for inputtable fields as they may confuse users to think they are disabled fields. All fields that users can interact with and put input should have a clear background e.g. all formula fields

prestonlimlianjie:
This PR fixes:
- the color of the valid input fields in the Calculated Result and Conditional Result logic components, and
- the map buttons in the Map Constants logic component. 

They were previously set to grey (which is used to indicate that the button is disabled).

#### Before fix:
Screenshot 2021-04-09 at 5 54 23 PM
Screenshot 2021-04-09 at 5 54 28 PM
Screenshot 2021-04-09 at 6 22 08 PM

#### After fix:
Screenshot 2021-04-09 at 5 47 51 PM
Screenshot 2021-04-09 at 5 47 47 PM
Screenshot 2021-04-09 at 6 20 17 PM

Fixes https://github.com/opengovsg/checkfirst/issues/333

### DIFF
--- a/src/client/components/builder/logic/CalculatedResult.tsx
+++ b/src/client/components/builder/logic/CalculatedResult.tsx
@@ -51,7 +51,6 @@ const InputComponent: OperationFieldComponent = ({ operation, index }) => {
           name="expression"
           type="text"
           placeholder="Enter expression"
-          bg="#F4F6F9"
           fontFamily="mono"
           value={expression}
           onChange={(expression) =>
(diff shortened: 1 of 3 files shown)


lamkeewei (approved):
lgtm!

PR #361, opened 2021-04-09
fix: make non-button components unclickable

### LINKED ISSUE #330: Remove unnecessary interactivity on Map Constants + IF/ELSE Logic component
### Problem
The `MAP` and `TO` elements of the `Map Constants` component are not interactive, and hence should not have a hover state/be clickable. Same for the `IF/ELSE` component. Issue raised by @NatMaeTan.

Screenshot 2021-04-06 at 2 45 56 PM

### Fix
Remove the highlight and ensure mouse does not change to mickey mouse glove on hover

prestonlimlianjie:
This PR makes the non-button components unclickable. Previously, the `IF`, `THEN`, `ELSE` buttons of the if/then and the `MAP` and `TO` buttons of the map logic components were clickable buttons.

#### Screenshots
Screenshot 2021-04-09 at 6 08 40 PM
Screenshot 2021-04-09 at 6 06 48 PM

Fixes https://github.com/opengovsg/checkfirst/issues/330

### DIFF
--- a/src/client/components/builder/logic/ConditionalResult.tsx
+++ b/src/client/components/builder/logic/ConditionalResult.tsx
@@ -6,6 +6,7 @@ import {
   Divider,
   Button,
   IconButton,
+  Heading,
   VStack,
   HStack,
   Box,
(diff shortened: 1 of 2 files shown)


lamkeewei (approved):
lgtm!

PR #368, opened 2021-04-12
fix: sort references based on order in question and logic tab

### LINKED ISSUE #334: Reference dropdown is not in order
### Problem
In dropdown, questions are not arranged in order. Issue raised by @NatMaeTan.
Screenshot 2021-04-06 at 2 57 44 PM

### Fix
It’s best if we can order it by the order in the questions tab

lamkeewei:
## Problem

Closes #334 

## Solution
**Improvements**:
- Added `baseSort` in match-sorter options that specifies a sort order based on the position in `items`
- Refactored `items` to use `useMemo` hook to prevent repeated evaluation

## Before & After Screenshots

image

image

## Tests
- Create a checker with multiple fields
- Create a calculated result field and enter `@` in the input
- Press down to trigger auto-completion
- The order should be based on the position of the fields in the question tab
- All fields should also appear before operations

Note: If you entered a query string after `@`, the sort order will be first based on the closeness of the match (defined [here](https://github.com/kentcdodds/match-sorter#this-solution)), followed by the position in the checker.

### DIFF
--- a/src/client/components/builder/logic/ExpressionInput.tsx
+++ b/src/client/components/builder/logic/ExpressionInput.tsx
@@ -1,4 +1,5 @@
-import React, { FC } from 'react'
+import React, { FC, useMemo } from 'react'
+import { findIndex } from 'lodash'
 import {
   HStack,
   VStack,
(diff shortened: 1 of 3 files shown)


src/client/components/builder/logic/ExpressionInput.tsx
```diff
@@ -105,6 +110,15 @@ export const ExpressionInput: FC<ExpressionInputProps> = ({
     return changes
   }
 
+  const baseSort = (
```
prestonlimlianjie, line 117:
```suggestion
// The baseSort function is used to tie-break items that have the same ranking.
// This is useful in two main cases:
// - when the user types `@ `, we want to display all reference options on the Questions tab followed by the Logic tab; these reference options should be sorted according to their position on their respective tabs.
// - when the user types `@ <query>` and there are multiple reference options with titles that exactly match `<query>`, we want to sort these reference options according to their position on their Questions tab followed by the Logic tab.
  const baseSort = (
```

prestonlimlianjie: we should leave a comment to briefly explain what `baseSort` does; please feel free to rephrase

prestonlimlianjie (approved):
lgtm - left a comment

PR #385, opened 2021-04-14
fix: instantiate sequelize within lambda handler

prestonlimlianjie:
### Problem
@lamkeewei spotted intermittent Sequelize connection errors in our AWS lambda logs. 

image (1)

We hypothesize that these errors are occurring because we currently initialize Sequelize outside of our lambda handler, i.e. in the `Init` phase of the [Lambda execution lifecycle](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-context.html). Refer to `api.ts` for more details.

As such, the Sequelize instance and its connections are reused across multiple invocations, and these connections occasionally become stale/expire, resulting in the errors we see above.

#### Current `api.ts`:
```
import serverless, { Handler } from 'serverless-http'

import bootstrap from '../bootstrap'
const apiHandler = bootstrap().then((app) => serverless(app))

export const handler: Handler = async (event, context) => {
  const h = await apiHandler
  return h(event, context)
}
```

### Fix
We came up with two possible fixes:
- move the Sequelize instantiation into the handler
- test if the current connection has expired - if it has, request for a new connection from the Sequelize pool

We went with the former because it was simpler to implement. However, we recognize that this could result in a longer lambda execution time because each lambda function needs to create a new Sequelize instance (and hence connection pool) with the DB. 

We will continue to update this PR with latency numbers.


lamkeewei:
From what I am seeing initially, there seems to be a bumped in execution time for the handler, as expected. The 50th percentile execution time increased from `129ms` to `376ms`, comparing with stats from the period between last week to yesterday.

Ran the following in CloudWatch Log insights to get the stats with different time ranges:
```
filter @type='REPORT'
| stats pct(@billedDuration, 50)
| sort @timestamp desc
```

But qualitatively, it feels about the same navigating through the application.

src/server/serverless/api.ts
```diff
@@ -1,9 +1,13 @@
...
 
 export const handler: Handler = async (event, context) => {
-  const h = await apiHandler
-  return h(event, context)
+  const { app, sequelize } = await bootstrap()
+  const h = serverless(app)
+  try {
+    return await h(event, context)
+  } finally {
+    await sequelize.close()
```
prestonlimlianjie, line 13: Should we add a `logger.info('Closed sequelize connection')` here?

LoneRifle: we could.. trying to avoid bringing it in as a module to depend on though

LoneRifle (approved):

lamkeewei:
Force pushed it because GitGuardian check was stuck.

PR #390, opened 2021-04-16
fix(types): fix types and resultant type errors

### LINKED ISSUE #340: Type alias returns `any` bug
Investigate why our type alias returns `any` instead of the specific type.
Screenshot 2021-04-06 at 4 28 36 PM

orbitalsqwib:
## Problem

Type aliases returns any instead of the specific type due to invalid Pick<> syntax being used (`,` was used to chain the fields we wanted to pick instead of `|`), which confused the TS compiler as there were too many arguments.

Closes #340

## Solution

Fix export types by replacing `,` with `|` and subsequently fix uncaught type errors within the codebase.

**Bug Fixes**:

- fix export types for checker and user .d.ts files
- fix Checker ORM Model typings
- fix related uncaught checker and auth type errors in checker-service and checker-controller

### DIFF
--- a/src/server/checker/CheckerController.ts
+++ b/src/server/checker/CheckerController.ts
@@ -50,6 +50,12 @@ export class CheckerController {
   get: (req: Request, res: Response) => Promise<void> = async (req, res) => {
     const { id } = req.params
     const { user } = req.session
+
+    if (!user) {
+      res.status(401).json({ message: 'User not signed in' })
+      return
+    }
+
     try {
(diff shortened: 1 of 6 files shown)


LoneRifle:
There seems to be an unrelated change in this PR concerning user authentication..

orbitalsqwib:
The change related to user authentication was made as I found another Pick<> syntactical error in `user.d.ts`, which thus lead to type errors in that field getting uncovered.

LoneRifle (approved):
Lgtm

PR #392, opened 2021-04-16
refactor(models): explicitly define timestamps in model class

### LINKED ISSUE #391: Unable to create type aliases that use `createdAt` or `updatedAt` from the model
This is because we have defined the timestamps using the [`@Table` annotation](https://github.com/RobinBuschmann/sequelize-typescript#table):
```
@Table({
  timestamps: true,
  ...
})
class Person extends Model {}
```

As such, the class interface definition does not include `updatedAt` and `createdAt`.

Instead, we should define these timestamp fields using their [explicit decorators](https://github.com/RobinBuschmann/sequelize-typescript#createdat-updatedat-deletedat) so that our class interfaces have these fields:
```
  @CreatedAt
  createdAt: Date;

  @UpdatedAt
  updatedAt: Date;
```

prestonlimlianjie:
## Problem

This PR explicitly defines timestamps in the model class so that we can create type aliases from these classes with the various timestamp fields.

Currently, we are unable to create - for example - a new type alias from `Checker` with an `updatedAt` field using the `Pick<>` utility. This is because the `Checker` class/interface does not have `updatedAt` explicitly defined in it.

Closes https://github.com/opengovsg/checkfirst/issues/391

## Solution

We explicitly define `updatedAt` and `createdAt` using their respective specific decorators (i.e. `@UpdatedAt` and `@CreatedAt`) within the model class.

### DIFF
--- a/src/server/database/models/Checker.ts
+++ b/src/server/database/models/Checker.ts
@@ -1,10 +1,12 @@
 import {
   BelongsToMany,
   Column,
+  CreatedAt,
   DataType,
   HasMany,
   Model,
   Table,
+  UpdatedAt,
 } from 'sequelize-typescript'
 
 import { Constant, Display, Field, Operation } from '../../../types/checker'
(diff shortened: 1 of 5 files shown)


LoneRifle (approved):
lgtm

Issue #367, opened 2021-04-12
Open edit checker page after "Create" checker


lisa-kwok:
**Is your feature request related to a problem? Please describe.**
Currently after clicking "Create" button for new checker, I will be brought to my dashboard with a list of checker. 
Since the order of the checker is not based on recency (which will be tackled in issue #366), it takes some time to look for the checker that a user has created.

**Describe the solution you'd like**
A smoother experience would be to bring the user directly to edit the checker he/she has just created instead of clicking into the new checker again.

PR #397, opened 2021-04-16
fix(create-new-modal): open builder upon successful checker creation

### LINKED ISSUE #364: New checkers with description not opening into editor
### **Problem**
When creating a new checker, if we only fill in the title field, checker opens into editor when clicking _Create_. 
However, if you also fill in a description, the screen returns back to the dashboard when clicking _Create_.

Screenshot 2021-04-12 at 2 18 20 PM

### **Fix**
Clicking _Create_ should always open into the editor.

orbitalsqwib:
## Problem

When creating a new checker, if we only fill in the title field, checker opens into editor when clicking Create.
However, if you also fill in a description, the screen returns back to the dashboard when clicking Create.

Additionally, since the order of the checker is not based on recency (which will be tackled in issue #366), it takes some time to look for the checker that a user has created.

Closes #364 and #367

## Solution

**Bug Fixes**:

- fix `CreateNewModal` to open the editor once a checker is successfully created

## Tests

- [x] Check that the builder is opened with the correct title/description for each test
- [x] Check that the builder is opened after creating a checker from scratch
- [x] Check that the builder is opened after creating a checker from a template
- [x] Check that the builder is opened after creating a checker when duplicating another checker

### DIFF
--- a/src/client/components/dashboard/CreateNewModal.tsx
+++ b/src/client/components/dashboard/CreateNewModal.tsx
@@ -1,5 +1,11 @@
 import React, { FC } from 'react'
-import { useParams, Switch, Redirect, Route } from 'react-router-dom'
+import {
+  useParams,
+  Switch,
+  Redirect,
+  Route,
+  useHistory,
+} from 'react-router-dom'
 import { useQueryClient, useQuery, useMutation } from 'react-query'
 import {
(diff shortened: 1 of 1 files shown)


lamkeewei (approved):
lgtm!

PR #398, opened 2021-04-19
fix(builder): select first operation by default

### LINKED ISSUE #388: Attempting field actions within the builder while active index is invalid results in a crash
**Describe the bug**
This bug is caused by attempting to execute field actions, such as moving a field up or down within the builder, while the active index is out of range of the current set of fields.

**To Reproduce**
1. Create a new form
2. Move to the logic section
3. Add any result
4. Note that this new builder field is not expanded, or "active"
5. Click on either the "Move up" or "Move down" buttons
6. App will handle crash by returning the "Whoops, something went wrong." page

Alternatively, deleting any first field while there are two or more fields in the section, then carrying out step 5 will also result in the same behavior being replicated.

**Expected behavior**
All field actions should be disabled while the active index is invalid. Additionally, there should be an active field at all times unless there are no fields in the section at all.

lamkeewei:
## Problem

`activeIndex` is initialised to `0` when there's no operation. After adding the first operation, `activeIndex` is incremented to `1`, which is an invalid index.

Closes #388 

## Solution

**Bug Fixes**:

- First operation should be selected by default if there is at least one operation

## Tests
- Create a new checker
- Create a new operation and observe the following:
   - Operation should be selected by default
   - Shift up/down should be disabled

### DIFF
--- a/src/client/components/builder/LogicTab.tsx
+++ b/src/client/components/builder/LogicTab.tsx
@@ -34,6 +34,7 @@ import {
   DateResult,
 } from '../builder/logic'
 import { BuilderActionEnum, ConfigArrayEnum } from '../../../util/enums'
+import useActiveIndex from '../../hooks/use-active-index'
 
 // Images
 import emptyLogicTabImage from '../../assets/states/empty-logic.svg'
@@ -74,9 +75,7 @@ const generateDefaultDateOp = (id: number): checker.Operation => ({
 
 export const LogicTab: FC = () => {
   const { dispatch, config } = useCheckerContext()
-  const [activeIndex, setActiveIndex] = useState<number>(() =>
-    config.operations.length > 0 ? 0 : -1
-  )
+  const [activeIndex, setActiveIndex] = useActiveIndex(config.operations)
   const [offsetTop, setOffsetTop] = useState<number>(16)
   const [nextUniqueId, setNextUniqueId] = useState<number>(1)
 
--- a/src/client/components/builder/LogicTab.tsx
+++ b/src/client/components/builder/LogicTab.tsx
@@ -73,10 +73,12 @@ const generateDefaultDateOp = (id: number): checker.Operation => ({
 })
 
 export const LogicTab: FC = () => {
-  const [activeIndex, setActiveIndex] = useState<number>(0)
+  const { dispatch, config } = useCheckerContext()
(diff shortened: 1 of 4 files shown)


src/client/components/builder/LogicTab.tsx
```diff
@@ -65,10 +65,12 @@ const generateDefaultDateOp = (id: number): checker.Operation => ({
 })
 
 export const LogicTab: FC = () => {
-  const [activeIndex, setActiveIndex] = useState<number>(0)
+  const { dispatch, config } = useCheckerContext()
+  const [activeIndex, setActiveIndex] = useState<number>(() =>
+    config.operations.length > 0 ? 0 : -1
```
prestonlimlianjie: There's still a bug with the `activeIndex` implementation where if we created 2 Logic tab components (where O1 is at index 0 and O2 is at index 1) and then proceeded to delete the first component (O1 at index 0), the `move down` button for the last remaining component is still active. And if we clicked on that `move down` button, an error will be thrown.

prestonlimlianjie (changes requested):

src/client/components/builder/LogicTab.tsx
prestonlimlianjie: This actually happens in all 3 tabs^

prestonlimlianjie: https://github.com/opengovsg/checkfirst/blob/9d0270b0b068767d27d22b4c61ea3120a27dc954/src/client/components/builder/BuilderField.tsx#L199-L213

I think we need to do a check in the `handleDelete` function in `BuilderField.tsx`, where we check both the length of the array corresponding to `configArrName` and current `index`. The current issue is that we don't pass the actual array into the `handleDelete` function, otherwise we could do:

#### Could be represented using a chain of ternary operators
```
index > 0 ? setActiveIndex(index - 1) : (configArr.length === 1 ? setActiveIndex(-1) : setActiveIndex(0))
```

#### Or nested if/else statements
```
    if (index === 0) {
      if (configArr.length === 1) {
        setActiveIndex(-1)
      } else {
        setActiveIndex(0)
      }
    } else {
      setActiveIndex(index - 1)
    }
```

lamkeewei: Thanks for catching this! Will fix.

prestonlimlianjie (approved):
lgtm!

PR #399, opened 2021-04-19
chore: migrate to husky v6 using cli

prestonlimlianjie:
This PR migrates from husky v4 to v6 using the CLI command described [here](https://github.com/typicode/husky-4-to-6#npm). I forgot to run it before merging this PR https://github.com/opengovsg/checkfirst/pull/384.

This PR moves the husky hooks' definitions out of `package.json` and into the `.husky` folder as recommended in https://typicode.github.io/husky/#/?id=automatic-recommended. We wanted to follow the docs/recommendations to set them up in a vanilla way so that subsequent developers can understand how we've set up husky without having to trawl through `package.json`.

## Note:
Added `npm test` in the `pre-push` hook


LoneRifle (approved):
lgtm

PR #407, opened 2021-04-20
feat(expression-input): improve query referencing behaviour

### LINKED ISSUE #336: Improve referencing behavior when the @ key is typed
### Problem
Currently nothing shows when the @ key is typed. Issue raised by @NatMaeTan.
Screenshot 2021-04-06 at 3 00 39 PM

### Fix
Ideally, the dropdown would appear immediately when the @ key is entered

orbitalsqwib:
## Problem

Currently nothing shows when the `@` key is typed. Additionally, query replacement can only occur with one `@` character at a time, and will treat all characters behind the `@` as part of the query, of which the latter is unintended behaviour.

Closes #336

## Solution

**Features**:

- allows query replacement with more than one `@`
- improves replacement behaviour 
- remove isOpen toggle from state reducer to ensure that caret pos is updated before checking for query strings

**Bug Fixes**:

- fixes destructive replacement where queries originally replaced all content behind it
- fixes down arrow revealing replacement dropdown when there are no queries to replace
- fixes replacement dropdown not showing upon `@` character being typed

## Tests

Note: `a|a` -> `aa|a` refers to typing the second a into the string, where `|` indicates the placement of the caret

- [ ] Check that replacement dropdown appears on typing `@`
- [ ] Check that replacement dropdown replaces the second `@` when typing `@ + | + @` -> `@ + @| + @`
- [ ] Check that the replacement dropdown does not appear when pressing the down arrow key with the expression `123`
- [ ] Check that replacing ` @ ` does not trim the whitespace behind it or after it

### DIFF
--- a/src/client/components/builder/logic/ExpressionInput.tsx
+++ b/src/client/components/builder/logic/ExpressionInput.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useMemo } from 'react'
+import React, { FC, useMemo, useState } from 'react'
 import { findIndex } from 'lodash'
 import {
   HStack,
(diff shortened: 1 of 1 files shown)


lamkeewei (approved):
lgtm! Great work, much more robust than my initial hacky implementation.

PR #412, opened 2021-04-20
chore: add compile job to catch compilation failures

lamkeewei:
## Problem

Compilation failures are not caught by our non-deploy CI workflow currently. This results in potential compilation failures to be missed until we attempt to deploy to staging.

## Solution

**Improvements**:
- Add job to run build to catch possible typescript/webpack compilation failures


prestonlimlianjie (approved):
lgtm!

PR #413, opened 2021-04-21
fix: add back scroll to results

lamkeewei:
Users seems to want the scroll to results behaviour we've removed previously. Adding this back in.


prestonlimlianjie (approved):
lgtm!

PR #424, opened 2021-04-21
fix(expression-query): fix caret behaviour on query replace

orbitalsqwib:
## Problem

Currently, replacing a query string sends the caret (text input cursor) to the end of the expression, and negatively affects the user experience when working with long expressions.

Fixes #407

## Solution

This solution fixes the unintended behaviour and sends the user's text cursor to just after the replaced value.

**Improvements**:

- update input caret selection after replace to after replaced value for better user experience when editing long expressions
- clean up state reducer, reduced replaceVariableName calls to 1 from 2

**Bug Fixes**:

- shift caretPos from `useState` to `useRef` to fix issues arising from asynchronous state updates

## Before & After

**BEFORE**:
`... > @myfield| + ...` ---- replace ----> `... > N1 + ...|`

**AFTER**:
`... > @myfield| + ...` ---- replace ----> `... > N1| + ...`

## Tests

- [x] Check that replacing a query string does not send the text input cursor to the end of the expression

### DIFF
--- a/src/client/components/builder/logic/ExpressionInput.tsx
+++ b/src/client/components/builder/logic/ExpressionInput.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useMemo, useState } from 'react'
+import React, { FC, useMemo, useRef } from 'react'
 import { findIndex } from 'lodash'
 import {
   HStack,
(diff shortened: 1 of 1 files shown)


lamkeewei (approved):
lgtm! Thanks for fixing!

PR #426, opened 2021-04-22
fix(checker): fix overscroll behaviour on embedded checkers

### LINKED ISSUE #421: iframe parent page scrolls with checkfirst page
### Problem
Here is an example Isomer page that iframes a checker page - when the user clicks on the `Submit` button, the Isomer page is scrolled along with the checker page.

https://user-images.githubusercontent.com/19917616/115515830-14b4af80-a2b8-11eb-8972-d39037cc76ff.mov

### Solution
Find a way to prevent scrolling on the parent page

orbitalsqwib:
## Problem

fix overscroll behaviour on embedded checkers

Closes #421

## Solution

**Features**:

- set `overscroll-behaviour` to `contain`

### DIFF
--- a/src/client/components/Checker.tsx
+++ b/src/client/components/Checker.tsx
@@ -149,7 +149,7 @@ export const Checker: FC<CheckerProps> = ({ config }) => {
 
   return (
     <StylesProvider value={styles}>
-      <Container maxW="xl" p={8} mb={4}>
+      <Container maxW="xl" p={8} mb={4} overscroll-behaviour="contain">
         <FormProvider {...methods}>
           <form onSubmit={methods.handleSubmit(onSubmit)}>
             <VStack align="stretch" spacing={10}>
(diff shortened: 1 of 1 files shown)


lamkeewei:
@orbitalsqwib it seems that just adding `overscroll-behavior` doesn't solve the problem.

https://user-images.githubusercontent.com/3666479/116049761-3e4f4b80-a6a9-11eb-8d67-923871d71396.mov

The parent still gets scrolled when `scrollIntoView` is called. This can be solved by adding `block: 'nearest'` as part of the options for `scrollIntoView`. Although `scrollIntoViewOptions` is non-standard and [does not work on every browser](https://caniuse.com/?search=scrollIntoViewOptions), it seems that the issue of scrolling the parent doesn't happen on Safari (both iOS and desktop). Scrolling to results just doesn't work on iOS for now. Will open a separate issue to address this.

lamkeewei (approved):

PR #471, opened 2021-04-30
feat(dropdown-field): implement combobox for dropdown fields

### LINKED ISSUE #425: allow text input/search on dropdown
This helps users with more than 10 dropdown items to easily search. e.g countries.
This behavior is also consistent with formsg's dropdown feature.

orbitalsqwib:
## Problem

Currently, the dropdown fields in CheckFirst do not have search capabilities. This is bad UX as users will have to scan long lists of options manually to find their desired option.

Closes #425

## Solution

This PR adds a reusable and performant combobox component to replace the current `select` dropdown for dropdown fields.

**Features**:

- Fast and snappy search behaviour
- Supports large lists, e.g > 1000 rows

## Before & After Screenshots

**BEFORE**:
Screenshot 2021-04-30 at 3 28 51 PM

**AFTER**:
Screenshot 2021-04-30 at 3 28 39 PM

## Tests

- [x] Check that search works
- [x] Check that clicking out of the input field resets the input field value to the last selected element
- [x] Check that clicking a row via mouse or arrow keys+enter sets the input field value to that row's label
- [x] Check that the correct value is returned by the combobox when referenced in a result field

## Deploy Notes

**New dependencies**:

- `react-window` : used to reduce the render lag on very large lists

**New dev dependencies**:

- `@types/react-window` : used to type `react-window`


src/client/components/common/Combobox.tsx
```diff
@@ -0,0 +1,358 @@
...
+              itemData={
+                {
+                  items: searchResults.current,
+                  highlightedIndex: highlightedIndex,
+                  getItemProps: getItemProps,
+                  selectedItem: selectedItem,
+                  topDropdownInset: dropdownInset,
+                } as ItemRendererProps['data']
+              }
+              innerElementType={innerElementType}
```
lamkeewei, line 342: Do we need to set this prop? Isn't the default value a `div` and also the styling is applied on the `ListItem`?

src/client/components/common/Combobox.tsx
```diff
@@ -0,0 +1,358 @@
...
+
+  return (
+    <>
+      <ListItem
+        {...getItemProps({
+          style: {
+            ...props.style,
+            // purely for adding padding insets to FixedStyleList
+            // see https://codesandbox.io/s/react-window-list-padding-dg0pq
+            top: `${parseFloat(`${props.style.top}`) + topDropdownInset}px`,
```
lamkeewei, line 68:
```suggestion
            top: `${parseFloat(`${props.style.top}`) + topDropdownInset}px`,
            listStyle: 'none',
```

Currently bullets are still shown for list items.

src/client/components/common/Combobox.tsx
```diff
@@ -0,0 +1,358 @@
...
+  // Avoid using refs for dropdown direction as it's only used
+  let dropdownDir = DropdownDirection.down
+
+  // Note that inputRef is a merged ref as we may not want to
+  // always provide a forward ref for the input
+  const fallbackInputRef = useRef<HTMLInputElement>(null)
+  const inputRef = inputOptions?.forwardRef || fallbackInputRef
+
+  const dropdownList = useRef<FixedSizeList & HTMLElement>(null)
+  const searchResults = useRef<ComboboxItem[]>(items)
```
lamkeewei: Is there as specific reason to use `useRef` instead of `useState`? 

Generally, `useRef` should be used for things that persists throughout the entire lifecycle of a component and `useState` is used for things that causes a re-render. Given that when the search results changes, a re-render would be necessary, might be more appropriate to use `useState`.

src/client/components/common/Combobox.tsx
```diff
@@ -0,0 +1,358 @@
...
+   * Calculates and returns array of the the items with the
+   * closest matching labels to the inputValue
+   *
+   * @param inputValue The user query to search the dropdown with.
+   * @returns An array of all matching items.
+   */
+  const getTopSearchResults = (inputValue: string) => {
+    let searchResults: ComboboxItem[]
+    if (inputValue === '') {
+      searchResults = items
```
lamkeewei: I think the [default behaviour](https://runkit.com/embed/gzr7mj0wkam1) for `matchSorter` is to return all items when `inputValue` is empty.

lamkeewei (changes requested):
Left some comments/questions! Two other general comments are:
1. Might want to look into the use of [useMultiStyleConfig](https://chakra-ui.com/docs/theming/component-style#styling-multipart-components) to centralise styles for the different parts of the combo box (input, dropdown item, etc).
2. Not sure if it might be worth refactoring to make use of `useCombobox` to simplify the code a little.

Great job on looking into the use of `react-window` to handle potentially huge lists.

src/client/components/common/Combobox.tsx
orbitalsqwib: This prop is required to add the 8px in-list padding at the top and bottom as required in the design, as the row elements use absolute positioning within the div and thus require us to manually extend the height of the div element to account for the padding.

orbitalsqwib:
Hi @lamkeewei, thanks for the helpful comments on this PR!

In regards to 1, I think I'll open a new issue to implement the style centralization and work on it on a separate PR. 

However, I think 2 might not be worth the refactor effort as it will affect the order in which the functions are called, thus essentially requiring large changes to the current code. Do let me know if this is something you feel strongly about and I'll try my best to implement it.

lamkeewei (approved):
lgtm!

PR #486, opened 2021-05-04
feat: set only one condition type for each ifelse component

### LINKED ISSUE #376: Disallowing mix of additional conditions for IF/ELSE
### **Problem**
In IF/ELSE, the additional condition dropdown appears for every _AND/ OR_ , which allows users to have a mix of both conditions. 

Screenshot 2021-04-14 at 9 59 09 AM

### **Fix**
If we are disallowing such a mix, we should only allow the dropdown for the very first _AND/OR_ and also auto update the other conditions to the new _AND/OR_ when it is changed.

prestonlimlianjie:
This PR restricts each IF/ELSE logic component to have only one condition type, i.e. if the user chooses the `AND` condition type, all conditions within the same IF/ELSE component must use the `AND` type.

The behavior is described within the following hi-fidelity screen: https://www.figma.com/file/qtoA63mFtCW3jGYhr8ev7y/Master?node-id=1420%3A13161

Fixes https://github.com/opengovsg/checkfirst/issues/376.

### DIFF
--- a/src/client/components/builder/logic/ConditionalResult.tsx
+++ b/src/client/components/builder/logic/ConditionalResult.tsx
@@ -1,6 +1,6 @@
 import React, { useState, useEffect } from 'react'
 import { isValidExpression, math } from '../../../core/evaluator'
-import update from 'immutability-helper'
+import update, { Spec } from 'immutability-helper'
 import {
   BiGitBranch,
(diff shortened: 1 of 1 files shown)


lamkeewei (approved):
lgtm!

PR #502, opened 2021-05-05
fix: set correct initial condition type for if/else

prestonlimlianjie:
The if/else component was setting the default initial condition type as `AND` - even though the existing if/else component has an `OR` condition type. This results in a bug where the preview component/expression correctly uses the `OR` condition type, while the input component uses the default `AND` condition type.

### Preview component
Screenshot 2021-05-05 at 3 42 58 PM

### Input component

Screenshot 2021-05-05 at 3 42 55 PM


lamkeewei (approved):
lgtm! Tested this locally!

PR #503, opened 2021-05-05
fix(default-tooltip): close tooltip on mouse down and click

### LINKED ISSUE #499: Frozen tooltip bug
**Describe the bug**
_Add condition_ tooltip doesn't disappear after non-hover and seems to be stuck in place

**To Reproduce**
Unsure what the pattern to reproduce is, just clicked on the icon multiple times
https://user-images.githubusercontent.com/63220614/117105648-1058be00-adb1-11eb-9e1d-ea237b63d12f.mov

**Expected behavior**
Tooltip should disappear when not hovering over.

**Screenshots**
Screenshot 2021-05-05 at 2 39 40 PM

**Additional context**
Unsure if this was worsened by the tooltip lag to display. The pause was intentional, but can be removed if it's confusing the system.

orbitalsqwib:
## Problem

Add condition tooltip doesn't disappear after non-hover and seems to be stuck in place

Closes #499

## Solution

Update default tooltip to close tooltips on mouse down and click

### DIFF
--- a/src/client/components/common/DefaultTooltip.tsx
+++ b/src/client/components/common/DefaultTooltip.tsx
@@ -7,7 +7,7 @@ import React, { FC } from 'react'
  * @see Docs     https://chakra-ui.com/components/tooltip
  */
 export const DefaultTooltip: FC<TooltipProps> = ({ ...props }) => (
-  <Tooltip {...props} />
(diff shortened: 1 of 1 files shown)


prestonlimlianjie (approved):
lgtm

PR #504, opened 2021-05-05
fix(expression-input): fix query string replacement inputvalues not updating correctly

orbitalsqwib:
## Problem

Whenever a query string replacement occurs, the input value is not updated correctly, which is unexpected behaviour for expression inputs.

## Solution

Update where onChange is called to after the state reducer is run to ensure that the latest state is captured for updates.


prestonlimlianjie (approved):
lgtm

PR #506, opened 2021-05-05
feat(app): implement code and bundle splitting

### LINKED ISSUE #430: Enable code-splitting in webpack
Currently, the entire codebase for the frontend is bundled into a single file. This presents two issues:
- The source map for the bundle is also in a single file. While not a problem in and of itself, the size of the file breaches response size [limits](https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html#:~:text=Payload%20size) in API Gateway
- Not all of the frontend has to be loaded at runtime. For example, members of the public will never see the admin dashboard, and public officers only need to see certain parts of the dashboard at certain points

This issue hence proposes the use of [code-splitting](https://reactjs.org/docs/code-splitting.html) to achieve the above two goals. Care should be taken to adjust the paths allowed by API Gateway to accommodate the new split code bundles.

orbitalsqwib:
## Problem

Currently, the entire codebase for the frontend is bundled into a single file. This presents two issues:

- The source map for the bundle is also in a single file. While not a problem in and of itself, the size of the file breaches response size limits in API Gateway
- Not all of the frontend has to be loaded at runtime. For example, members of the public will never see the admin dashboard, and public officers only need to see certain parts of the dashboard at certain points

Closes #430

## Solution

Use code-splitting to achieve the above two goals and adjust the paths allowed by API Gateway to accommodate the new split code bundles.

**Features**:

- implement bundle splitting in webpack
- implement code splitting in app.tsx using @loadable/component
- update bundle export paths to `build/client/assets/js`
- update API Gateway paths to accommodate new bundles
- update page exports to export their components by default

## Tests

- [x] Check that the main bundle is split into main, vendor, page and library chunk bundles.

## Deploy Notes

**New dependencies**:

- `@loadable/component` : manages lazy loading of components and covers edge cases that `React.lazy` does not

**New dev dependencies**:

- `@types/loadable__component` : provides types for `@loadable/component`


lamkeewei:
@orbitalsqwib could you elaborate a little more about what edge cases `loadable` handles that `React.lazy` doesn't.

orbitalsqwib:
Hi @lamkeewei,
The edge cases that `loadable` can handle as opposed to `React.lazy` are that it can handle server-side rendering, library splitting to defer the loading of a library until required and full dynamic imports which allow for aggressive code splitting by Webpack.

I mainly used `loadable` as it was recommended by [this](https://reactrouter.com/web/guides/code-splitting) react router article on code splitting and did not require nesting the lazy loaded components deeper within a React `Suspense` element, which I feel makes the code cleaner and easier to maintain.

LoneRifle:
> library splitting to defer the loading of a library

@orbitalsqwib I didn't really see this in action; the checker view still loaded the full vendor bundle. is this something we need to follow up on?

orbitalsqwib:
No, this implementation doesn't use the full feature set of `loadable`, though it would be useful if we wanted to lazy load more sections of the application.

lamkeewei (approved):
lgtm! Tested this on staging and works as expected. 

This has pushed the source map size to ~9MB. The longer term solution to fix this would still be to ensure that Sentry source maps upload works as expected. Currently, the files uploaded are not correct.

PR #508, opened 2021-05-06
fix(expression-input): hotfix to fix onchange trigger

orbitalsqwib:
## Problem

Currently, Downshift's `onInputValueChange` event handler runs before the state reducer is able to update the input value, thus returning the wrong value to the `onChange` callback function.

## Solution

Shift the `onChange` callback to Downshift's `onStateChange` event handler to ensure that the state reducer will always be run before `onChange` is called.

Additionally, refactor the state reducer to remove the constant declaration for `newInputValue` and shift the calculation to within the return value.

**Bug Fixes**:

- Calls `onChange` with the correct expression value

## Tests

- [ ] Check that replacing `@` -> `@DL1`, then clicking out of the builder field, displays the preview as `DL1` (Standard query replacement)
- [ ] Check that replacing `ABC + @` -> `ABC + DL1`, then clicking out of the builder field, displays the preview as `ABC + DL1` (Standard query replacement with other characters)
- [ ] Check that typing `ABC + DL1` -> `ABC + DL1@`, then clicking out of the builder field, displays the preview as `ABC + DL1@` (No query replacement, adding other characters)
- [ ] Check that any other changes made to the expression input is preserved after clicking out of the builder field (any other edge cases you can think of)

### DIFF
--- a/src/client/components/builder/logic/ExpressionInput.tsx
+++ b/src/client/components/builder/logic/ExpressionInput.tsx
@@ -123,16 +123,14 @@ export const ExpressionInput: FC<ExpressionInputProps> = ({
   ) => {
     switch (changes.type) {
       case Downshift.stateChangeTypes.clickItem:
-      case Downshift.stateChangeTypes.keyDownEnter: {
-        const newInputValue = replaceVariableName(
-          state?.inputValue,
-          changes?.inputValue
-        )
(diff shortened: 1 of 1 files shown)


lamkeewei:
@orbitalsqwib what is the user impact of this issue?

lamkeewei (approved):
lgtm!

PR #514, opened 2021-05-11
fix(otpform): lowercase email before sending otp

### LINKED ISSUE #513: enable log in of emails with all cases
when someone logs in using email formatted in all capital letters, it doesn't work.

Screenshot 2021-05-11 at 9 45 17 AM

**desired behaviour**
It should either still log in the user, or it shows an error message to remind users to not type their email in caps.

orbitalsqwib:
## Problem

When someone logs in using email formatted in all capital letters, it doesn't work.

Closes #513

## Solution

Currently, the user model email column and minimatch checks on the domain force emails to be in lowercase. To allow uppercase emails to be used during the sign in procedure, we can lowercase the input strings before passing it to the auth API. 

**Bug Fixes**:

- Format email to lowercase before sending it through to auth APIs.

## Tests

- [ ] Check that logging in works with uppercased / mixed case emails.

### DIFF
--- a/src/client/components/OtpForm.tsx
+++ b/src/client/components/OtpForm.tsx
@@ -28,7 +28,11 @@ export const OtpForm: FC<OtpFormProps> = ({ onSuccess }) => {
   const onSubmit = (data: { email: string }) => {
     sendOtp.reset()
     const { email } = data
-    sendOtp.mutate(email)
+
+    // format email to database expectations
+    const formattedEmail = email.toLowerCase()
+
+    sendOtp.mutate(formattedEmail)
   }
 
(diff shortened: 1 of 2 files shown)


lamkeewei (changes requested):
@orbitalsqwib generally it's better to do this sort of transformation/validation on the backend to ensure that it will work in all cases. Related to [Postel's Law](https://en.m.wikipedia.org/wiki/Robustness_principle).

lamkeewei (approved):
lgtm! I've also checked existing production database for any prior user email addresses that are mixed-case to prevent potentially creating a new user since all emails are lowercased now. None were found.

PR #516, opened 2021-05-11
fix(form-builder): fix scroll to invalid input behaviour 

### LINKED ISSUE #515: improve scroll to invalid input behaviour on preview and published checkers
**Describe the bug**

After submission of a form with missing or invalid fields, the checker will currently automatically focus on the offending field (and by extension, scroll it into view).

However, the preview nav bar is not taken into account when this occurs during the preview modal, and thus displays a different behaviour than expected, as the focused element is often hidden behind the nav bar.

This can lead to confusion on the user's part as to why the checker is not submitting as expected, and can be misconstrued as a bug.

Additionally, on published checkers, only the input of the offending field is scrolled into view, excluding the title and description. This requires users to scroll up slightly more to see what is the context behind the invalid field.

**To Reproduce**
1. Create a long checker that extends beyond a screen's length
2. Edit a field at the top of the checker to be invalid, and it should not be seen when the submission button is in view
3. Click on the submission button. It should now focus and scroll to the invalid field.
4. Note that the invalid field and its associated details are obscured by the nav bar.

**Expected behavior**
The invalid field element should be scrolled into view, including the title and description.

**Screenshots**
Scroll to invalid input behaviour on a published checker. Note that the title is out of the frame.
Screenshot 2021-05-11 at 11 28 26 AM

Scroll to invalid input behaviour on a previewed checker. Note that the input is obscured by the nav bar.
Screenshot 2021-05-11 at 11 28 39 AM

Scroll to invalid input behaviour on a previewed checker, after scrolling up a bit more.
Screenshot 2021-05-11 at 11 28 47 AM

orbitalsqwib:
## Problem

After submission of a form with missing or invalid fields, the checker will currently automatically focus on the offending field (and by extension, scroll it into view).

However, the preview nav bar is not taken into account when this occurs during the preview modal, and thus displays a different behaviour than expected, as the focused element is often hidden behind the nav bar.

This can lead to confusion on the user's part as to why the checker is not submitting as expected, and can be misconstrued as a bug.

Additionally, on published checkers, only the input of the offending field is scrolled into view, excluding the title and description. This requires users to scroll up slightly more to see what is the context behind the invalid field.

Closes #515

## Solution

**Features**:

- implemented virtual controller inputs to support `react-hook-form` scroll into view for irregular, non-standard inputs that consist of more than one input element, such as radio groups, checkbox fields and date selectors.

**Improvements**:

- Add a scroll container over the tab content, such that nav bars are not included as part of the scroll area.
- Add top scroll margins to virtual / non-virtual inputs to include field titles and descriptions on a best effort basis

## Before & After Screenshots

**BEFORE**:
Previewed checker
Screenshot 2021-05-11 at 11 28 39 AM

Published checker
Screenshot 2021-05-11 at 11 28 26 AM

**AFTER**:
Previewed checker
Screenshot 2021-05-11 at 12 02 34 PM

Published checker
Screenshot 2021-05-11 at 12 02 57 PM

## Tests

- [ ] Check that published checker scrolls to invalid inputs as per above screenshots (title and description both visible)
- [ ] Check that previewed checker scrolls to invalid inputs as per above screenshots (should have exactly the same behaviour as published checkers)


lamkeewei:
Was wondering if there's a simpler approach to solving this problem than to introduce a `VirtualControllerInput`. Not sure if there's any unintended side effects if the `ref` is not bounded to the actual input.

orbitalsqwib:
According to the react-hook-form documentation, the `field.ref` return value from `Controller` is only responsible for focusing the error inputs.

> A ref used to connect hook form to the input. Assign ref to component's input ref to allow hook form to focus the error input.

As such, I don't think there should be other side effects from assigning the ref to a virtual input, other than that it might disrupt navigation/accessibility.

I think I can give fixing the refs on problematic fields such as `DateField` and `RadioField` another shot, but if it doesn't pan out, the virtual input might be better alternative than having a broken scroll-to-focus

lamkeewei:
> According to the react-hook-form documentation, the `field.ref` return value from `Controller` is only responsible for focusing the error inputs.
> 
> > A ref used to connect hook form to the input. Assign ref to component's input ref to allow hook form to focus the error input.
> 
> As such, I don't think there should be other side effects from assigning the ref to a virtual input, other than that it might disrupt navigation/accessibility.
> 
> I think I can give fixing the refs on problematic fields such as `DateField` and `RadioField` another shot, but if it doesn't pan out, the virtual input might be better alternative than having a broken scroll-to-focus

Sure - I think worthwhile to take a second look. Just felt a little uneasy fixing a minor UI issue with a fairly non-standard change.

orbitalsqwib:
Managed to fix the refs across the troublesome fields. Turned out to be issues with how Chakra composes and styles the Radio and Checkbox inputs, as well as how Flatpickr manages the refs on the rendered input field.

Main notable changes are the two additional css files that add `scrollMarginTop` to the internal inputs on the Chakra Radio/Checkbox fields, which should be okay for our current use case, provided that we don't need specific scroll margins for each field.

src/client/styles/checkbox-field-input.css
```diff
@@ -0,0 +1,3 @@
+.chakra-checkbox__input {
+    scroll-margin-top: 88px;
```
lamkeewei: Is it possible to override this value [using Chakra](https://chakra-ui.com/docs/theming/component-style#styling-single-part-components) instead of patching the CSS directly? How Chakra name their CSS classes should be transparent to us.

orbitalsqwib: It might be possible to override the value through overriding the global default input themes, but I feel that this solution might be too invasive and cause unexpected behaviours on modals where we may want to customise the same set of styles on another page with different values.

As such, I'd still prefer that we just override the CSS directly, though we could discuss about this if you have strong opinions on the subject.

lamkeewei: As discussed on Slack, this current way of overriding the CSS also affects _all_ checkbox inputs since it's overriding global styles.

lamkeewei (approved):
lgtm!

PR #522, opened 2021-05-11
fix: set correct active ifelse condition menu item

### LINKED ISSUE #500: If/else logic component and/or should not automatically select first menu item
Currently, we use the default `autoSelect=true` behavior as prescribed by [Chakra UI's menu component](https://chakra-ui.com/docs/overlay/menu#usage). However, this could result in confusion - take for example, this If/else component which has the **OR** condition type selected. When a user clicks on the condition type toggle, the first element of the menu list is selected, making it seem like the **AND** condition type had been selected when it isn't the case.

Screenshot 2021-05-05 at 2 39 40 PM

Ideally, we would want the correct selected condition type to be highlighted.

prestonlimlianjie:
Fixes https://github.com/opengovsg/checkfirst/issues/500.

### Currently-selected condition type is set to active
Screenshot 2021-05-11 at 12 25 20 PM

### User hovers mouse on non-currently-selected condition type
Screenshot 2021-05-11 at 12 25 25 PM

### DIFF
--- a/src/client/components/builder/logic/ConditionalResult.tsx
+++ b/src/client/components/builder/logic/ConditionalResult.tsx
@@ -240,7 +240,7 @@ const InputComponent: OperationFieldComponent = ({ operation, index }) => {
         </HStack>
         {ifelseState.conditions.map((cond, i) => (
           <HStack key={i}>
-            <Menu>
+            <Menu autoSelect={false}>
               {i === 0 ? (
                 <Box w="100px">
                   <MenuButton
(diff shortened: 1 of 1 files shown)


prestonlimlianjie:
Note: @NatMaeTan pointed out that the fix introduces a new blue box shadow. This currently occurs because the fix uses a `Box` to set a default `isActive` state. We discussed and decided that the easiest thing to do is to wait till the design system style refactor is done to fix the box shadow at the source.

lamkeewei (approved):
lgtm! Tested and works as expected!

PR #523, opened 2021-05-11
fix(dropdown-field): fix combobox behaviours to match user expectations

### LINKED ISSUE #520: improve combobox behaviours to match user expectations
**Describe the bug**

Currently, the combobox selects the first option's value by default. However, this is not good UX as users may inadvertently forget to update the field with their own selection and thus get unexpected results.

**Expected behavior**
The combobox should have no default value, and instead display a placeholder to prompt the user to select a value themselves. If the user submits the checker without having filled in a combobox, it should highlight the combobox as invalid and scroll the field into view.

Finally, the clear button should set the combobox to an empty value, instead of resetting only the search value. Hence, when clicking out or losing focus, the combobox should not revert to a valid option from the menu, remaining blank.

orbitalsqwib:
## Problem

Currently, the combobox selects the first option's value by default. However, this is not good UX as users may inadvertently forget to update the field with their own selection and thus get unexpected results.

Closes #520

## Solution

**Improvements**:

- The default value of dropdown fields are now empty
- Comboboxes display 'Select an option' as a placeholder when no option has been selected
- The combobox clear button now unsets any selected option (if any) on top of clearing the input element
- Focusing on the combobox no longer clears the input element

**Bug Fixes**

- Fixed combobox returning a falsy `0` value instead of a string `'0'` on combobox changes, which affects users if the first option was chosen

## Example

https://user-images.githubusercontent.com/21305518/117758576-e2fa8d00-b254-11eb-83c8-924597fae537.mov

## Tests

- [ ] Check that the combobox behaves as described / shown above

### DIFF
--- a/src/client/components/common/Combobox.tsx
+++ b/src/client/components/common/Combobox.tsx
@@ -197,13 +197,6 @@ export const Combobox: FC<ComboboxProps> = ({
     changes: StateChangeOptions<ComboboxItem>
   ) => {
     switch (changes.type) {
-      case Downshift.stateChangeTypes.clickItem:
-      case Downshift.stateChangeTypes.keyDownEnter:
-      case Downshift.stateChangeTypes.changeInput:
-        // recalculate top search results and scroll back to top of list
-        updateSearchResults(changes.inputValue || '')
-        dropdownListRef.current?.scrollTo(0)
-        break
       case Downshift.stateChangeTypes.blurInput:
       case Downshift.stateChangeTypes.mouseUp:
         // reset inputValue to last valid input value
@@ -237,6 +230,11 @@ export const Combobox: FC<ComboboxProps> = ({
       itemToString={(item) => item?.label || ''}
       onChange={(item) => {
         onChange(`${item ? item.value : ''}`)
+      }}
+      onInputValueChange={(inputValue) => {
+        // recalculate top search results and scroll back to top of list
+        updateSearchResults(inputValue)
+        dropdownListRef.current?.scrollTo(0)
       }}
       defaultHighlightedIndex={0}
     >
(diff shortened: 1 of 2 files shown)


src/client/components/common/Combobox.tsx
```diff
@@ -197,6 +197,8 @@ export const Combobox: FC<ComboboxProps> = ({
     changes: StateChangeOptions<ComboboxItem>
   ) => {
     switch (changes.type) {
+      case Downshift.stateChangeTypes.clickItem:
+      case Downshift.stateChangeTypes.keyDownEnter:
       case Downshift.stateChangeTypes.changeInput:
         // recalculate top search results and scroll back to top of list
         updateSearchResults(changes.inputValue || '')
```
lamkeewei: Might want to move this out of the `stateReducer` to keep it a pure function.

src/client/components/common/Combobox.tsx
```diff
@@ -280,12 +279,9 @@ export const Combobox: FC<ComboboxProps> = ({
                 }
                 {...getInputProps({
                   ...props,
+                  placeholder: 'Select an option',
                   ref: inputRef,
                   onFocus: () => {
-                    // reset field for new search
```
lamkeewei, line 285: I think we should keep this behaviour actually. After a user has already selected an option, re-clicking the option would only show 1 item in the dropdown. It's non-obvious that the user has to first clear their combo box. 

Form's dropdown behaviour is also to clear the previously selected item, so it will be good to remain consistent.

lamkeewei (changes requested):

src/client/components/common/Combobox.tsx
orbitalsqwib: I've discussed this with Natalie already, and it seems that she wants this behaviour as it might be closer to what the user expects with an input field. If users complain about this we can always revert back to the original clear input on focus implementation.

lamkeewei (approved):
lgtm! Thanks for the fix!

PR #539, opened 2021-05-12
fix: allow selection of title field when there is more than 1 question

lamkeewei:
## Problem
By default `setActiveIndex` only allow setting the index to `-1` if there is only one item in the collection (e.g. when deleting the last item). However, in question tab we make use of `-1` as a special index to represent the TitleField. As such, if there is more than one question, `setActiveIndex` will not set the `activeIndex` to `-1` to select TitleField. 

## Solution

**Bug Fixes**:

- Add a `force` flag to allow explicitly setting of the `activeIndex` to `-1`. Admittedly, this is not a long term fix and we should refactor to clean up the index management for each tab.

## Test
- [ ] Ensure that TitleField can be selected even if there is more than one question
- [ ] Ensure that shift up and shift down are disabled appropriately on the other tabs

### DIFF
--- a/src/client/components/builder/QuestionsTab.tsx
+++ b/src/client/components/builder/QuestionsTab.tsx
@@ -1,4 +1,5 @@
 import React, { FC, useEffect, useState } from 'react'
+import { noop } from 'lodash'
 import {
   BiHash,
   BiRadioCircleMarked,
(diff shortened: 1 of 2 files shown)


orbitalsqwib (approved):
lgtm

PR #612, opened 2021-05-25
feat(expression-input): implement basic calculator bar

### LINKED ISSUE #472: Add calculator bar to logic blocks
### **Problem**
1. Feedback from users that they are unsure how to type in math symbols
2. Unsure how to reference questions

### **Fix**
Include a calculator symbol bar into **calculation logic** and **conditional logic** blocks (It is unnecessary in other blocks)
Calculator bar helps users visualise/ see what math symbols they can use, as well as identify the reference symbol

Screenshot 2021-05-03 at 12 19 14 PM
Screenshot 2021-05-03 at 12 19 08 PM

[Figma](https://www.figma.com/file/qtoA63mFtCW3jGYhr8ev7y/Master?node-id=1969%3A27)

Behaviour:
- Calculation bar should always be present when **calculation logic** and **conditional logic** blocks are focused

orbitalsqwib:
## Problem

Users are unsure how to type in math symbols or how to reference questions, which is affecting reach and retention. To solve this, a calculator bar can help users visualize and see what math symbols they can use, as well as identify the reference symbol

Closes #472

## Solution

Implement a basic calculator bar for expression inputs, displayed on input focus.

**Features**:

- implement calculator bar into expression inputs and enable them to add/replace content within the input.

## Screenshots

Screenshot 2021-05-25 at 9 50 22 PM
Screenshot 2021-05-25 at 9 50 41 PM

## Tests

- [ ] Check that calculator bar can add operators into the expression input
- [ ] Check that entering the `@` operand through the calculator bar brings up the query replacement dropdown
- [ ] Check that selecting a selection of characters in the expression input, then subsequently clicking on a operand in the calculator bar will replace that selection with the new character (select and replace)


src/client/components/builder/logic/CalculatorBar.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface CalculatorBarProps extends Omit<FlexProps, 'onClick'> {
+  onClick: (character: string) => void
+}
+
+export const CalculatorBar: FC<CalculatorBarProps> = ({
+  onClick,
+  ...props
+}) => {
+  const operands = '@+-×÷()^%>≥<≤=≠'.split('')
```
prestonlimlianjie: We should create a map of operator to mappedOperator so that we won't run the risk of messing up the indices when we add/remove more operators in the future, e.g.
```
const operatorMap = { 
  "@": "@",
  "=": "==",
   ...
} 
```

P.S. it should be **operator** not **operand** - **operand** refers to the object of the operation, i.e. in `3 + 6`, `3` and `6` are the **operands** while `+` is the **operator**

src/client/components/builder/logic/ExpressionInput.tsx
```diff
@@ -186,93 +226,127 @@ export const ExpressionInput: FC<ExpressionInputProps> = ({
     return inputStr || ''
   }
 
+  const handleCalculatorChange = (operand: string) => {
```
prestonlimlianjie, line 208: Should be **operator** not **operand**

src/client/components/builder/logic/ExpressionInput.tsx
```diff
@@ -51,9 +52,48 @@ export const ExpressionInput: FC<ExpressionInputProps> = ({
   ...props
 }) => {
   const [inputValue, setInputValue] = useState<string>(value)
+  const [showCalcBar, setShowCalcBar] = useState(false)
   const [selection, setSelection] = useState<{ start: number; end: number }>()
 
   const inputRef = useRef<HTMLInputElement>(null)
+  const wrapperRef = useRef<HTMLDivElement>(null)
+
+  // syncs input value with subsequent value updates
```
prestonlimlianjie: The code in lines 61-80 is duplicated from line 98 onwards

prestonlimlianjie (changes requested):
The PR looks good - some minor changes requested below

src/client/components/builder/logic/ExpressionInput.tsx
orbitalsqwib: Thanks for catching this! I must've been tired when I was resolving merge conflicts with develop...

orbitalsqwib:
I've implemented the map as a Map instead of an Object so we can use iterators (and keep the order consistent without having to repeat ourselves with an array).

Let me know if this is okay with you!

prestonlimlianjie (approved):
lgtm!

PR #633, opened 2021-06-01
fix(frontend): parsing of conditional expressions with `and` or `or`

### LINKED ISSUE #630: Builder crash when conditional input has `and` or `or` in condition
**Describe the bug**
If the expression includes `and` or `or` within the condition, the application will crash as the parsing thinks that it is an invalid expression. For e.g. `R1 == "Singaporean or Permanent Resident" will crash the application.

lamkeewei:
## Problem

Closes #630 

## Solution

**Improvements**:
- Parse conditional expression to `IfelseState` using `mathjs` expression tree instead of regex
- Add test coverage for parsing to prevent regressions

**Bug fixes**
- Fix parsing for expression that includes `and` or `or`.
- Fix parsing for expression that includes parentheses. E.g. `R1 == "A (B)`.

## Tests
- [ ] Create a conditional input and insert an expression that compares against value that contains `and` or `or`. E.g. `R1 == "Singaporean or PR"`. The application should not crash.
- [ ] - [ ] Create a conditional input and insert an expression that compares against value that contains parentheses. E.g. `R1 == "A (B)"`. Clicking out and then back in to the block. The expression should be displayed correctly.

### DIFF
--- a/src/client/components/builder/logic/ConditionalResult.tsx
+++ b/src/client/components/builder/logic/ConditionalResult.tsx
@@ -1,5 +1,11 @@
 import React, { useState, useEffect } from 'react'
-import { isValidExpression, math } from '../../../core/evaluator'
+import {
+  ConditionType,
+  Condition,
+  IfelseState,
+} from '../../../../types/conditional'
+import { parseConditionalExpr } from '../../../core/parser'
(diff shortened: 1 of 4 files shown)


orbitalsqwib (approved):
lgtm

Issue #549, opened 2021-05-17
[Design System] Builder blocks


lamkeewei:
**Work required**
- Introduce colour highlights based on builder block type
- Align styles with design system. Introduce builder block variants style as necessary

Refer to [CheckFirst Design System](https://www.figma.com/file/X3oi4JnynZuJ2j6zSHDnTH/CheckFirst-Design-System) for details.

![image](https://user-images.githubusercontent.com/3666479/118431789-b5f12300-b709-11eb-9d4c-2b7baa986833.png)

PR #650, opened 2021-06-03
feat(builder-field): align with design system and introduce dynamic colour highlights

orbitalsqwib:
## Problem

Updates the `BuilderField` component to align with the new design system.

Part of #549

## Solution

**Features**:

- Border highlight colours change depending on the builder field's type

**Improvements**:

- Organize builder styles into its own individual folder
- Declare types for builder styles
- Implement `colorScheme` support for builder field components
- Refactored `ActionButton` styles as a separate component style
- Separate common field component styles such as `dummyInput` away from the main builder field component styles for clarity
- Aligned builder field layout and styles with design system

## Before & After Screenshots

**BEFORE**:
Screenshot 2021-06-03 at 4 32 26 PM

**AFTER**:
Screenshot 2021-06-03 at 4 31 36 PM

## Tests

- [ ] Check that dynamic color highlighting works as per design system (question tab - teal/green, constants tab - yellow, logic tab - dark blue)
- [ ] Verify that the new layout aligns with the design system (excluding input component/ preview components)

### DIFF
--- a/src/client/components/builder/BuilderField.tsx
+++ b/src/client/components/builder/BuilderField.tsx
@@ -102,7 +102,16 @@ export const createBuilderField =
     const [ref, { top }] = usePosition()
     const { dispatch } = useCheckerContext()
     const variant = active ? 'active' : ''
-    const styles = useMultiStyleConfig('BuilderField', { variant })
+    const colorScheme = !active
+      ? 'inactive'
+      : isFieldData(data) || isTitleData(data)
+      ? 'success'
+      : isConstantData(data)
+      ? 'warning'
+      : isOperationData(data)
+      ? 'primary'
+      : 'primary' // fallback
+    const styles = useMultiStyleConfig('BuilderField', { variant, colorScheme })
 
     useEffect(() => {
       if (active) onActive({ top })
@@ -240,7 +249,7 @@ export const createBuilderField =
           {(isFieldData(data) ||
             isOperationData(data) ||
             isConstantData(data)) && (
-            <Button colorScheme="primary" sx={styles.badge}>
+            <Button colorScheme={colorScheme} sx={styles.badge}>
               {data.id}
             </Button>
           )}
--- a/src/client/components/builder/BuilderField.tsx
+++ b/src/client/components/builder/BuilderField.tsx
@@ -102,15 +102,15 @@ export const createBuilderField =
     const [ref, { top }] = usePosition()
     const { dispatch } = useCheckerContext()
     const variant = active ? 'active' : ''
-    const colorScheme = !active
-      ? 'inactive'
-      : isFieldData(data) || isTitleData(data)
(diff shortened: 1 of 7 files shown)


src/client/theme/components/builder/BuilderField.tsx
```diff
@@ -0,0 +1,76 @@
+import { ComponentMultiStyleConfig } from '@chakra-ui/theme'
+
+const CommonComponents: ComponentMultiStyleConfig = {
```
lamkeewei, line 3: Any particular reason why we are splitting out `CommonComponents`?

src/client/theme/foundations/colors.ts
```diff
@@ -122,4 +122,9 @@ export const colors = {
     100: '#EDEDED',
     50: '#F9F9F9',
   },
+  inactive: {
```
lamkeewei, line 125: I think this can be removed. Don't think it's used anywhere.

src/client/components/builder/BuilderField.tsx
```diff
@@ -102,7 +102,16 @@ export const createBuilderField =
     const [ref, { top }] = usePosition()
     const { dispatch } = useCheckerContext()
     const variant = active ? 'active' : ''
-    const styles = useMultiStyleConfig('BuilderField', { variant })
+    const colorScheme = !active
```
lamkeewei: Might want to simplify this to make it clearer. The chained ternary expression is a bit hard to read.

```typescript
    const getVariant = (data: BuilderFieldData): string => {
      if (!active) return 'inactive'
      if (isFieldData(data) || isTitleData(data)) return 'success'
      if (isConstantData(data)) return 'warning'
      if (isOperationData(data)) return 'primary'
      return 'primary'
    }
    const colorScheme = getVariant(data)
```

lamkeewei (approved):
lgtm! Just some minor nits.

src/client/theme/foundations/colors.ts
orbitalsqwib: This color scheme is used for the builder block badges, and provides the proper color scheme for the unfocused states. That said, please do recommend a better name that might make this more obvious if you have one in mind, thanks!

src/client/theme/components/builder/BuilderField.tsx
orbitalsqwib: I decided to split these parts away from the `BuilderField` style declarations as they are more applicable to its child components. As such, it should make the `BuilderField` component styles easier to maintain without having to scan through a bunch of shared component styles that may not be applicable to the current edit.

src/client/theme/foundations/colors.ts
lamkeewei: Missed this out! Yep, we can keep this.

PR #653, opened 2021-06-04
feat(question-fields): align with design system and implement textarea for description inputs

orbitalsqwib:
## Problem

Updates the `TitleField`, `NumericField`, `RadioField`, `DropdownField`, `CheckboxField` and `DateField` components to align with the new design system.

Part of #549

## Solution

**Features**:

- Implement textarea for description inputs, including supporting newline characters
- Implement index number badges for question fields
- Implement big radio / checkboxes for radio and checkbox fields

**Improvements**:

- Organize question field styles into its own individual folder
- Align question field layout and styles with design system

## Before & After Screenshots

**BEFORE**:
Screenshot 2021-06-04 at 4 54 55 PM

**AFTER**:
Screenshot 2021-06-04 at 4 54 28 PM

## Tests

- [ ] Implement textarea for description inputs, including supporting newline characters
  - [ ] Check that the design lines up with the design system
  - [ ] Check that input works and changes are dispatched correctly
  - [ ] Check that newlines work in the field preview and checker preview
  - [ ] Check that the description is center aligned on the checker preview when it can fit on one line
  - [ ] Check that the description is left aligned on the checker preview when it takes up more than one line (or has a newline)
- [ ] Implement index number badges for question fields
  - [ ] Check that the index numbers are correct on a form (1-indexed)
  - [ ] Check that shifting fields up and down update the index numbers correctly
- [ ] Implement big radio / checkboxes for radio and checkbox fields
  - [ ] Check that the design lines up with the design system
  - [ ] Check that the radio / checkboxes checked states match the increased sizes correctly
- [ ] Verify that the new layout aligns with the design system (scope: question tab)

### DIFF
--- a/src/client/components/Checker.tsx
+++ b/src/client/components/Checker.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useState, useRef, useLayoutEffect } from 'react'
+import React, { FC, useState, useRef } from 'react'
 import { isEmpty, filter } from 'lodash'
 import {
   useToast,
@@ -48,19 +48,6 @@ export const Checker: FC<CheckerProps> = ({ config }) => {
   const [variables, setVariables] = useState<checker.VariableResults>({})
   const googleAnalytics = useGoogleAnalytics()
   const outcomes = useRef<HTMLDivElement | null>(null)
-
-  // conditionally toggle description label between center align (1 line) and left align (> 1 line)
-  const descriptionRef = useRef<HTMLParagraphElement>(null)
-  useLayoutEffect(() => {
-    if (descriptionRef.current && styles.subtitle.lineHeight) {
-      const height = descriptionRef.current.clientHeight
-      const singleLineHeight = parseInt(
-        styles.subtitle.lineHeight.toString().slice(0, -2), // slice removes 'px' suffix
-        10
-      )
-      if (height > singleLineHeight) descriptionRef.current.style.width = '100%'
-    }
-  }, [descriptionRef, description, styles.subtitle.lineHeight])
 
   const renderField = (field: checker.Field, i: number) => {
     switch (field.type) {
@@ -174,11 +161,7 @@ export const Checker: FC<CheckerProps> = ({ config }) => {
             <VStack align="stretch" spacing={10}>
               <VStack spacing={2}>
                 <Text sx={styles.title}>{title}</Text>
-                {description && (
-                  <Text sx={styles.subtitle} ref={descriptionRef}>
-                    {description}
-                  </Text>
-                )}
+                {description && <Text sx={styles.subtitle}>{description}</Text>}
               </VStack>
               {fields.map(renderField)}
               <Button colorScheme="primary" width="100%" type="submit">
--- a/src/client/components/Checker.tsx
+++ b/src/client/components/Checker.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useState, useRef } from 'react'
+import React, { FC, useState, useRef, useLayoutEffect } from 'react'
 import { isEmpty, filter } from 'lodash'
 import {
   useToast,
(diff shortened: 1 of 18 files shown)


src/client/components/Checker.tsx
```diff
@@ -49,6 +49,19 @@ export const Checker: FC<CheckerProps> = ({ config }) => {
   const googleAnalytics = useGoogleAnalytics()
   const outcomes = useRef<HTMLDivElement | null>(null)
 
+  // conditionally toggle description label between center align (1 line) and left align (> 1 line)
+  const descriptionRef = useRef<HTMLParagraphElement>(null)
```
lamkeewei, line 53: Could we just set `maxWidth` to `100%` for subtitle with no `text-align`? If the text is less than one line, the width will be auto to fit the content. If it is more than 1 line, it will be width `100%`.

image

lamkeewei: Could we just set `maxWidth` to `100%` for subtitle with no `text-align`? If the text is less than one line, the width will be auto to fit the content. If it is more than 1 line, it will be width `100%`.

image

src/client/components/builder/questions/RadioField.tsx
```diff
@@ -101,60 +111,76 @@ const InputComponent: QuestionFieldComponent = ({ field, index }) => {
...
+        )}
       </VStack>
       <RadioGroup>
-        <VStack alignItems="left" spacing={2}>
+        <VStack sx={styles.previewOptionsContainer} spacing={0}>
           {options.map(({ value, label }, i) => (
-            <Radio key={i} value={value} isChecked={false}>
-              {label}
-            </Radio>
+            <Flex sx={styles.previewOptionRowContainer}>
```
lamkeewei:
```suggestion
            <Flex key={i} sx={styles.previewOptionRowContainer}>
```

Key should be on the top-level element to avoid warning from React.

src/client/components/builder/questions/CheckboxField.tsx
```diff
@@ -101,60 +117,77 @@ const InputComponent: QuestionFieldComponent = ({ field, index }) => {
...
+        )}
       </VStack>
       <CheckboxGroup>
-        <VStack alignItems="left" spacing={2}>
+        <VStack sx={styles.previewOptionsContainer} spacing={0}>
           {options.map(({ value, label }, i) => (
-            <Checkbox key={i} value={value} isChecked={false}>
-              {label}
-            </Checkbox>
+            <Flex sx={styles.previewOptionRowContainer}>
```
lamkeewei:
```suggestion
            <Flex key={i} sx={styles.previewOptionRowContainer}>
```

lamkeewei (approved):
lgtm! minor nits.

src/client/components/Checker.tsx
orbitalsqwib: Hi, the reason why I had to opt for this particular method of toggling the width was because of newline characters, which could wrap the subtitle onto a new line without necessarily increasing the width to 100%. As such, it would wrap in weird ways like this:

Screenshot 2021-06-07 at 8 35 25 PM

As such, this snippet would ensure that the width would always be increased to 100% if the subtitle exceeds one line. However, I do feel that this function is slightly overkill for this sort of formatting so let me know if you have a cleaner solution for this, thanks!

orbitalsqwib:
I tried using a new workflow to fix minor issues with `git commit --fixup` and `git rebase -i --autosquash` and have fixed the radio and checkbox field key issues with this, do lmk if this is better or worse for you when you review, thanks!

src/client/components/Checker.tsx
lamkeewei: I think this is quite an edge case though. Might be more consistent to just use the default CSS layout. I have not tested, but I think if we resize the window, this wouldn't re-evaluate. Would prefer that we not handle this for now. We can always revisit this if users raise an issue about this rather than pre-optimising for this. The solution could also be a design change rather than a code one.

PR #666, opened 2021-06-08
feat(logic-fields): align styles to design system

orbitalsqwib:
## Problem

Update the `CalculatedField`, `ConditionalField`, `MapResult`, `DateResult` and `FormulaPreview` component styles to align with the new design system.

Part of #549

## Solution

**Improvements**:

- updated text styles
- added shared styles for expression inputs
- updated and simplified layout
- added folder for logic field specific styles
- added additional field-specific styles in a separate theme file

## Before & After Screenshots

**BEFORE**:
Screenshot 2021-06-08 at 11 57 02 AM

**AFTER**:
Screenshot 2021-06-08 at 11 53 48 AM

## Tests

- [ ] Check that the new component styles line up with the design system

### DIFF
--- a/src/client/components/builder/BuilderField.tsx
+++ b/src/client/components/builder/BuilderField.tsx
@@ -7,6 +7,8 @@ import {
   Flex,
   HStack,
   Button,
+  Text,
+  Spacer,
 } from '@chakra-ui/react'
 
 import { useCheckerContext } from '../../contexts'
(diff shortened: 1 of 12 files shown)


lamkeewei (changes requested):
Two minor issues:
1. Upon selection of variables in MapResult or DateResult, the dropdown does not close
2. The selected item in the dropdown for both of these components should be highlighted

lamkeewei (approved):
lgtm!

PR #681, opened 2021-06-09
feat(styled-toasts): align toasts to design system, extract styles

### LINKED ISSUE #547: [Design System] Toasts
**Work required**
- Implement a [custom Toast component](https://chakra-ui.com/docs/feedback/toast#custom-component) that follows design system styling

Refer to [CheckFirst Design System](https://www.figma.com/file/X3oi4JnynZuJ2j6zSHDnTH/CheckFirst-Design-System) for details.

orbitalsqwib:
## Problem

Align toast usages to the design system via a custom component.

Closes #547

## Solution

Implement a custom functional component to be used within a toast's render prop. However, due to the limitations of Chakra's toast implementation, it is not possible to capture the native props passed into the toast function call (e.g `title`/`description`), thus the functional component itself has to specify its own arguments.

Note: The icons do not match up with the design system entirely. This is because react-icons does not currently provide a way to obtain the solid icon variants for BoxIcons. This should be looked into on a separate issue, as it may entail a larger scope of modifications to the application than what is expected within this PR.

**Features**:

- implemented custom toast functional component
- implemented theme styles

**Improvements**:

- updated usages of toasts across application (omitted some redundant titles as well to align with design system)

## Before & After Screenshots

**BEFORE**:
Screenshot 2021-06-09 at 2 36 17 PM
Screenshot 2021-06-09 at 2 36 41 PM
Screenshot 2021-06-09 at 2 36 03 PM

Position: bottom-right corner

**AFTER**:
Screenshot 2021-06-09 at 2 41 22 PM
Screenshot 2021-06-09 at 2 42 23 PM
Screenshot 2021-06-09 at 2 42 41 PM

Position: top of screen, centered

## Tests

- [ ] Check that toasts are displayed in the correct position (top) and styled as according to the design system
- [ ] Check that toasts are dismissed after 6 seconds
- [ ] Check that toasts can be dismissed manually by clicking on the 'X' button.

### DIFF
--- a/src/client/components/common/StyledToast.tsx
+++ b/src/client/components/common/StyledToast.tsx
@@ -1,0 +1,71 @@
+import React, { FC } from 'react'
+import {
+  Box,
+  Flex,
+  HStack,
+  Icon,
+  IconButton,
+  RenderProps,
+  Spacer,
+  Text,
+  useMultiStyleConfig,
+  useToast,
+} from '@chakra-ui/react'
+import { BiCheckCircle, BiErrorCircle, BiX } from 'react-icons/bi'
+
+interface StyledToastProps extends RenderProps {
+  message: string
+  status: 'error' | 'success' | 'warning'
+}
+
+export const StyledToast: FC<StyledToastProps> = ({
+  message,
+  status,
+  onClose,
+}: StyledToastProps) => {
+  const styles = useMultiStyleConfig('StyledToast', { colorScheme: status })
+
+  const getIcon = (status: 'error' | 'success' | 'warning') => {
+    if (status === 'error' || status === 'warning') return BiErrorCircle
+    if (status === 'success') return BiCheckCircle
+    return BiCheckCircle
+  }
+
+  return (
+    <Flex
+      sx={styles.boxContainer}
+      role="alert"
+      aria-live="assertive"
+      aria-atomic="true"
+    >
+      <Box sx={styles.toastBox}>
+        <HStack spacing={2} sx={styles.container}>
+          <Icon
+            as={getIcon(status)}
+            sx={styles.icon}
+            title={`${status}-icon`}
+          />
+          <Text sx={styles.message}>{message}</Text>
+          <Spacer />
+          <IconButton
+            icon={<BiX />}
+            sx={styles.closeButton}
+            aria-label="close"
+            aria-hidden
+            variant="link"
+            onClick={onClose}
+          />
+        </HStack>
+      </Box>
+    </Flex>
+  )
+}
+
+export const useStyledToast = () => {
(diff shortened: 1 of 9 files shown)


orbitalsqwib:
Additional note for reviewers: Note that this PR introduces slight copy changes. I chose not to include the titles for most toast usages as they were redundant, but they may be preferred in cases where clarity is important. If you have any concerns on this point, please highlight any instances where you think the title may still be useful within the toast's message text and comment accordingly, thank you!

src/client/components/common/StyledToast.tsx
```diff
@@ -0,0 +1,71 @@
...
+        </HStack>
+      </Box>
+    </Flex>
+  )
+}
+
+export const useStyledToast = () => {
+  const toast = useToast({
+    position: 'top',
+    duration: 6000,
```
lamkeewei, line 81: Perhaps we can have the `render` function as part of this and just keep the same interface as the original `useToast`? `useStyledToast` should take in `UseToastOptions`. This way, the users of a styled toast do not need to be aware of the actual component they are using.

lamkeewei (changes requested):
One minor comment. lgtm otherwise.

src/client/components/common/StyledToast.tsx
orbitalsqwib: Ordinarily, I would have used the `render` function to hide the custom rendered component, as suggested above. However, I had to resort to this particular method as the render props provided through the render function do not include the title, description or status data passed into the toast prop, which means that we have to specify our own props through the custom component.

orbitalsqwib: Note that the use of a custom toast component also excludes us from using a fair bit of the Chakra Toast API (as the custom render component can only recieve the `id` and `onClose` render props), so this is something we might have to bear with if we want to implement our own custom toast modals.

lamkeewei: How about something like this:

```ts
export const useStyledToast = (
  options?: UseToastOptions
): ReturnType<typeof useToast> => {
  const toast = useToast({
    position: 'top',
    duration: 6000,
    ...options,
  })

  const toastImpl = (opts?: UseToastOptions) => {
    const status = opts?.status ?? 'info'
    const description = opts?.description ?? ''

    return toast({
      render: (props) => (
        <StyledToast status={status} message={description} {...props} />
      ),
    })
  }

  toastImpl.close = toast.close
  toastImpl.closeAll = toast.closeAll
  toastImpl.isActive = toast.isActive
  toastImpl.update = toast.update

  return toastImpl
}
```

lamkeewei (approved):
lgtm!

PR #682, opened 2021-06-09
fix(checker): improve field null checks on checker submission

### LINKED ISSUE #676: Numeric field does not recognise 0 as valid value
**Describe the bug**
When the value of a numeric field is 0, a not found error is thrown when evaluating an expression.

**To Reproduce**
1. Create a new numeric field
2. Use numeric field in a calculated result
3. Go to preview and submit (leaving value as 0)
4. Error should be thrown

orbitalsqwib:
## Problem

Improve null check within checker submission method to prevent erroneous coercion of `0` as a falsy value.

Closes #676

## Solution

**Bug Fixes**:

- improve null guard conditional in checker submission method

## Tests

1. Create a new numeric field
2. Use numeric field in a calculated result
3. Go to preview and submit (leaving value as 0)

- [ ] Check that submission works as expected, where the calculated field should use 0 as the result value.

### DIFF
--- a/src/client/components/Checker.tsx
+++ b/src/client/components/Checker.tsx
@@ -96,7 +96,7 @@ export const Checker: FC<CheckerProps> = ({ config }) => {
 
     fields.forEach((field) => {
       const { id, type, options } = field
-      if (!inputs[id]) return
+      if (inputs[id] === null || inputs[id] === undefined) return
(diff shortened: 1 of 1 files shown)


lamkeewei (approved):
lgtm!

PR #683, opened 2021-06-10
feat(line-display): add support for in-string hyperlink

### LINKED ISSUE #667: render hyperlink in results
this is an intermediate solution to allow rendering of hyperlink on results section. the long term plan is to allow buttons in the results section, this would be further explored in the results tab (#306)
currently all forms of texts are displayed as pure text. 

this feature has been requested by a few users since most of them would like to redirect the public to another page after the computation of the checker/calculator

enable users to input hyperlinks in the form of expression e.g link("Hello","https://google.com")

orbitalsqwib:
## Problem

Enable users to input hyperlinks in the form of expression e.g `link('Hello','https://google.com')`

Closes #667

## Solution

Implement in-string hyperlinks (e.g: `"Hello this is a link('link','https://google.com') that will redirect you to Google!"` via a link parser that operates on string results before being displayed.

However, do note that due to the nature of how double quotes `"` are used to denote strings, the text within the link function should either have no quotes (as in `link(link, https://google.com)` or single quotes (as above).

**Features**:

- add link parser to line-display

**Improvements**:

- updated theme styles for link icons

## Screenshots

Screenshot 2021-06-10 at 1 45 38 PM

## Tests

- [ ] Check normal use case: `"Hello link('Google', 'https://google.com')!"`
- [ ] Check use case without quotes: `"Hello link(Google, https://google.com)!"`
- [ ] Check use case with extra spaces: `"Hello link(    Google , https://google.com      )!"`
- [ ] Check use case with incomplete URL: `"Hello link(Google, google.com)!"`
- [ ] Check that clicking any of the above links directs you to Google on a new tab

### DIFF
--- a/src/client/components/displays/LineDisplay.tsx
+++ b/src/client/components/displays/LineDisplay.tsx
@@ -1,5 +1,6 @@
-import React, { FC } from 'react'
-import { Stack, Text, useMultiStyleConfig } from '@chakra-ui/react'
+import React, { FC, ReactNodeArray } from 'react'
+import { Icon, Link, Stack, Text, useMultiStyleConfig } from '@chakra-ui/react'
+import { BiLinkExternal } from 'react-icons/bi'
 
 // If the value is over this threshold, the result will be rendered as horizontal
 // regardless of whether it is in mobile view.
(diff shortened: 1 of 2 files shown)


lamkeewei:
@orbitalsqwib how about implementing `link` as a math.js function that expands to an `a` tag? After which before displaying on `LineDisplay`, we can escape the generated content with [xss](https://github.com/leizongmin/js-xss). Two concerns I have is:
1. I'm slightly reluctant to use our hand-rolled parsing. As exemplified by the if-else block, there might be a bunch of edge cases that we may miss
2. This issue might be compounded once we have more "display" types of functions, since this will introduce edge cases that relates to each possible permutation. For e.g. `link(image('https://file.go.gov.sg/test.png'), 'google.com')`. Implementing it as a math.js functions allows us to rely on math.js to handle the order of evaluation/operator precedence.

orbitalsqwib:
> @orbitalsqwib how about implementing `link` as a math.js function that expands to an `a` tag? After which before displaying on `LineDisplay`, we can escape the generated content with [xss](https://github.com/leizongmin/js-xss). Two concerns I have is:
> 
>     1. I'm slightly reluctant to use our hand-rolled parsing. As exemplified by the if-else block, there might be a bunch of edge cases that we may miss
> 
>     2. This issue might be compounded once we have more "display" types of functions, since this will introduce edge cases that relates to each possible permutation. For e.g. `link(image('https://file.go.gov.sg/test.png'), 'google.com')`. Implementing it as a math.js functions allows us to rely on math.js to handle the order of evaluation/operator precedence.

Discussed and tested outside of Github, verdict was that using `mathjs` to return HTML links that would be rendered within a  React component through `dangerouslySetInnerHTML` would expose too many attack surfaces to sanitize against. Henceforth, this issue will be put on hold for now pending further discussion with the team.

lisa-kwok:
can we use other more laymen methods - such as allowing the rendering of html hyperlinks? for example when someone insert <a href="url">link text</a>, we'll just render it as a hyperlink?

lamkeewei:
Superseded by #805

PR #689, opened 2021-06-14
feat: add cache headers and hash to filename for static assets

### LINKED ISSUE #644: Add hashes to generated bundle files
After each deploy, the browser might not load the latest version of the `js` assets as the filename remains the same and a cached copy would be loaded instead. However, after a deploy this may result in errors as the old code in the user's browser request for now invalid/modified chunks.

Add an appropriate hashes (chunkhash, hash) to filenames to ensure that browser always load a valid set of files.

lamkeewei:
## Problem
Apart from #644, this PR also addresses a others issues that we were having with our caching configuration previously:
1. CloudFlare was always revalidating static assets instead of caching it as `Cache-Control` was set to `public,max-age=0`.
2. Static assets were not getting revalidated correctly because the `Last-Modified` header was always set to a date in 1980. Express by default uses the last modified date of the file on the filesystem. But in this case, running in lambda result in an unexpected behaviour.
3. Request for non-existent assets were returning `index.html` instead of returning a 404. This results in CloudFlare caching `index.html` as the response for any non-existent routes

Closes #644 

## Solution
- Set last modified date to the deployed timestamp so that requests can be revalidated correctly
- Added a CloudFlare page rule to address (1)
- Only return `index.html` for only defined routes
- Set `no-store` for all pages serving `index.html` to ensure that they are never stored

## Tests
- [ ] First deploy staging branch and load the page. Allow the use of cache by unchecking the `Disable cache` option on the Network tab of Chrome inspector. Deploy this branch and refresh page. You should see that the new js files are requested.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,7 @@ jobs:
             echo OTP_SECRET=${{ secrets.OTP_SECRET_STAGING }} >> $GITHUB_ENV;
             echo GA_TRACKING_ID=${{ secrets.GA_TRACKING_ID_STAGING }} >> $GITHUB_ENV;
             echo APP_HOST=staging.checkfirst.gov.sg >> $GITHUB_ENV;
+            echo DEPLOY_TIMESTAMP=$(date) >> $GITHUB_ENV;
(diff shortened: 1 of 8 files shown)


orbitalsqwib (approved):
lgtm

PR #694, opened 2021-06-15
fix(title-preview-text): fix preview title wrap behaviour

orbitalsqwib:
## Problem

The field index component is currently aligned to the center of the builder block title whenever it wraps onto more than one line. Instead, it should stay as an inline element on the first title line, with subsequent title lines wrapping below it as shown below.

Screenshot 2021-06-14 at 2 30 15 PM

## Solution

**Features**:

- use field index text as inline element
- consolidate title preview elements into a single component for clarity

**Improvements**:

- update shared styles to refactor out preview title styles
- updated usages of field index text across builder fields

**Bug Fixes**:

- fix preview title layout in new component

## Before & After Screenshots

See screenshot above

## Tests

- [ ] Check that title text wraps as expected, shown in the screenshot above.


lamkeewei (approved):
lgtm!

PR #702, opened 2021-06-15
fix(design-system): issues from design review

lamkeewei:
## Problem
Addressed review issues listed at [here](https://docs.google.com/document/d/1a21Ufg-_Y-VziyajJQN1FnxMijOYvE8iNswxz278GaE/edit#).

## Solution
**Bug Fixes**
- Add spacing after the last block for each builder tab
- Fix color for reference cube to `#FFFFFF`
- Standardise all border radius to `4px`
- Set placement for all tooltips to `right`
- Match icons for dropdown question field and toolbar
- Match icons for checkbox question field and toolbar
- Change tooltip to "Delete row" for deleting entry in mapping table
- Change button label to "Add condition" for conditional results


orbitalsqwib (approved):
lgtm

PR #712, opened 2021-06-16
fix(default-tooltip): remove tooltip delay

orbitalsqwib:
## Problem

Currently, the tooltip delay has a tendency to create visual glitches, which is turning out to affect the UI more than it actually helps it. As such, the designer has indicated that we should remove the tooltip delay to keep the UI smooth and glitch-free rather than deal with the intermittent glitches.

## Solution

**Bug Fixes**:

- Remove the 800ms open delay from `DefaultTooltip`

## Tests

- [x] Tooltips should now open instantly on hover


lamkeewei (approved):
lgtm!

Issue #393, opened 2021-04-16
Creating extremely long checker titles result in multiple UI issues


orbitalsqwib:
**Describe the bug**
Checker titles do not have an upper bound on length and are also not truncated when exceeding view bounds. This results in unexpected UI behavior when displaying titles in areas such as the checker duplication modal, toasts and builder modals.

**To Reproduce**
1. Create a checker with a very, very long title.
2. Go into the builder

**Expected behavior**
Titles should be truncated once exceeding a certain length or when they exceed a component's view bounds. Alternatively, they should be validated to have an upper bound placed on their length.

**Screenshots**
<img width="1792" alt="Screenshot 2021-04-16 at 2 48 12 PM" src="https://user-images.githubusercontent.com/21305518/114983133-cfad0980-9ec2-11eb-9fd5-910adf7419f8.png">
<img width="1792" alt="Screenshot 2021-04-16 at 2 47 26 PM" src="https://user-images.githubusercontent.com/21305518/114983150-d50a5400-9ec2-11eb-9be0-5d6c3d29156d.png">
<img width="1792" alt="Screenshot 2021-04-16 at 2 47 11 PM" src="https://user-images.githubusercontent.com/21305518/114983154-d5a2ea80-9ec2-11eb-9400-f17c1e2fd15e.png">

PR #715, opened 2021-06-17
fix(styled-toast): fix text overflow behaviour using styles

orbitalsqwib:
## Problem

Currently, styled toasts don't handle messages longer than itself properly. To align with the design system, toasts should be a maximum of one line long on desktop screens and two lines long on mobile screens.

Partially solves #393

## Solution

**Bug Fixes**:

- Update message styles to implement text cutoff after the maximum length is reached

## Before & After Screenshots

**BEFORE**:
Screenshot 2021-06-17 at 5 16 51 PM

**AFTER**:
Screenshot 2021-06-17 at 5 17 32 PM

## Tests

- [ ] Create a checker with a long title. Ensure that the toast does not exceed one line on desktop, or two lines on mobile.

### DIFF
--- a/src/client/theme/components/StyledToast.tsx
+++ b/src/client/theme/components/StyledToast.tsx
@@ -30,7 +30,7 @@ export const StyledToast: ComponentMultiStyleConfig = {
       pointerEvents: 'auto',
     },
     container: {
-      alignItems: 'stretch',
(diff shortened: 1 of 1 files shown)


lamkeewei (approved):
lgtm!

PR #775, opened 2021-07-07
feat: collaborator flow

### LINKED ISSUE #266: collaborators on checkers
Checkers are likely to be maintained by multiple users just like forms. 

Implementation:
1. Checker owner will be the admin
2. Other users must be added by the checker owner
3. Optional: should we allow changing of owners? this means checker transfer. 

Additional considerations: how to handle concurrency issues?

rlyss:
### Problem
Closes #266 

### Solution
Implemented 

- Frontend Settings Modal
     - Toggle checker active
     - Share links
     - Managing collaborators

- Backend migrations
     - Add isActive column to checkers table
     - Add isOwner column to usersToCheckers table
     - Backend routes for collaborators

image

### Behaviour
- [x]  All published checkers should remain published and active
- [x]  Turning a checker to inactive would hide it from the public
- [x]  Publishing an unpublished checker should also make it active immediately
- [x]  After the first publish, if checker has been set to inactive afterwards, publishing should not activate the checker automatically
- [x]  Newly created checkers start off as inactive
- [x]  Checkers without published checkers cannot be activated
- [x]  Share link redirect button should be disabled if checker inactive
- [x]  Error toast if added collaborator is not a user yet
- [x]  User cannot add themselves as a collaborator
- [x]  Returns to dashboard if collaborator deletes themselves
- [x]  Added collaborators should be able to access and edit checker
- [x]  Removed collaborators should no longer be able to access checker
- [x]  Inactive checker should be displayed on the dashboard

### Note
- Do not rollback migration `20210629030633-modify-users-to-checkers-association-add-owners-column.js` in the event of a rollback since all users regardless of checker ownership should maintain access to the checker and isOwner info must be preserved for migrations back up.


.sequelizerc
```diff
@@ -1,7 +1,7 @@
 const path  = require('path')
 
 module.exports = {
-    "config": path.resolve('./build/server/database/config', 'config.js'),
+    "config": path.resolve('./build/server/database/config', 'config.json'),
```
aniruddha-adhikary: Just curious, should this remain `config.js`? Didn't notice anything regarding the change of configs on the PR overview page. 👀

aniruddha-adhikary (commented):
Just a little possible development environment creep-up that caught my eye.

.sequelizerc
lamkeewei: Good catch @aniruddha-adhikary! @rlyss this should be `config.js`.

src/server/checker/__test__/CheckerController.spec.ts
```diff
@@ -29,6 +30,9 @@ describe('CheckerController', () => {
     delete: jest.fn(),
     retrievePublished: jest.fn(),
     publish: jest.fn(),
+    listCollaborators: jest.fn(),
```
lamkeewei, line 33: Let's add tests for the new controller methods too.

.sequelizerc
rlyss: Yup thanks for the reminder! I changed it to test out some stuff in sqlite and was going to restore it back to js in the next commit.

src/server/checker/CheckerService.ts
```diff
@@ -92,6 +100,10 @@ export class CheckerService {
   ) => Promise<GetPublishedCheckerWithoutDraftCheckerDTO | null> = async (
     id
   ) => {
+    // Check if checker is active
+    const checker = await this.CheckerModel.findByPk(id, {
```
lamkeewei: Given that now we allow multiple users to access/modify the same checker, we should wrap operations consecutive SQL operations up in a transaction to ensure a consistent results. In this case, for the initial publishing of a checker, there might be a situation where you return `isActive` is `false` but also yet return a published checker.

src/server/checker/CheckerService.ts
```diff
@@ -164,6 +181,97 @@ export class CheckerService {
...
+  ) => Promise<void> = async (id, user, collaboratorEmail) => {
+    const transaction = await this.sequelize.transaction()
+    const options = this.isSqliteFile ? {} : { transaction }
+
+    try {
+      const checker = await this.findAndCheckAuth(id, user)
+      collaboratorEmail = collaboratorEmail.toLowerCase()
+
+      if (checker) {
+        const collaboratorUser = await this.UserModel.findOne({
```
lamkeewei, line 279: Same for this, this should be wrapped in a transaction. Otherwise, you might end up with SQL constraint errors when there are concurrent requests trying to add the same user.

src/server/checker/CheckerService.ts
```diff
@@ -164,6 +181,97 @@ export class CheckerService {
...
+  deleteCollaborator: (
+    id: string,
+    user: User,
+    collaboratorEmail: string
+  ) => Promise<void> = async (id, user, collaboratorEmail) => {
+    const checker = await this.findAndCheckAuth(id, user)
+    if (checker) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      const collaborator = checker.users?.find(
```
lamkeewei: If `removeUser` returns a value, we can potentially use that to determine if the collaborator exists in the first place. Otherwise, similarly, we should wrap this up in a transaction.

lamkeewei: If `removeUser` returns a value, we can potentially use that to determine if the collaborator exists in the first place. Otherwise, similarly, we should wrap this up in a transaction.

lamkeewei (changes requested):
Did an initial pass of the backend. Generally, we should try to wrap SQL operations up in a transaction to avoid potential race conditions/inconsistencies. This is more so a case now that we are allowing multiple collaborators for a single checker. I've raised a few potential areas where this might be a problem. Do review the other parts of the code too.

Also as a side note, we should generally avoid the use of `ts-ignore`. If there's really no other option, do leave a comment justifying for its use. Overuse defeats the purpose of using a language that has type checking.

src/server/database/migrations/20210629030633-modify-users-to-checkers-association-add-owners-column.js
```diff
@@ -0,0 +1,20 @@
...
+      defaultValue: true,
+    })
+  },
+
+  down: async (queryInterface, _) => {
+    // return queryInterface
+    //   .bulkDelete('usersToCheckers', { isOwner: false })
+    //   .then(() => {
+    //     queryInterface.removeColumn('usersToCheckers', 'isOwner')
+    //   })
```
barnettx: to delete?

lamkeewei: Good catch! @rlyss just to be clear, in the event of a rollback, the behaviour we want is such that all users (regardless of owner or not) would preserve access to the checker right? Do try to test this scenario out also.

src/client/components/dashboard/SettingsModal.tsx
```diff
@@ -0,0 +1,412 @@
...
+                  government email may be added as collaborators
+                </Text>
+              </FormLabel>
+              <HStack spacing={2}>
+                <Input
+                  placeholder="me@example.gov.sg"
+                  value={inputEmail}
+                  onChange={handleInputChange}
+                />
+                <Button
```
lamkeewei, line 408: Disabled the button when the value for `inputEmail` is empty. You can also add the `isLoading` prop when the request is being processed to give immediate feedback to the user after the button is clicked.

image

src/client/components/dashboard/SettingsModal.tsx
```diff
@@ -0,0 +1,412 @@
...
+            <Text color="grey">Owner</Text>
+          ) : (
+            <Text>Editor</Text>
+          )}
+        </Td>
+        <Td px={0} py={0}>
+          {collaborator.UserToChecker.isOwner ? (
+            <Box height="40px" />
+          ) : (
+            <IconButton
```
lamkeewei, line 72: I think the delete option should be disabled for the logged in user. Otherwise, it is a bit strange to delete himself as a collaborator.

image

src/client/components/dashboard/SettingsModal.tsx
```diff
@@ -0,0 +1,412 @@
...
+          </Text>
+        </FormLabel>
+      )
+    }
+  }
+
+  return (
+    <Modal isOpen onClose={onClose} size="xl">
+      <ModalOverlay />
+      <ModalContent py="16px">
```
lamkeewei, line 311: Maybe we can add a loading spinner while `publishedChecker` and `collaborators` are being fetched to prevent the UI from flashing between states. Can consider using `isFetched` from `react-query`.

lamkeewei (changes requested):
Some minor UI changes.

src/client/components/dashboard/SettingsModal.tsx
rlyss: Added checker loading states

https://user-images.githubusercontent.com/44181699/126972467-4b3cfaa7-9aca-4036-b68b-f4c1fe0971a7.mov

src/server/checker/CheckerService.ts
```diff
@@ -164,6 +190,114 @@ export class CheckerService {
...
+
+      if (checker) {
+        const collaboratorUser = await this.UserModel.findOne({
+          where: { email: collaboratorEmail },
+          ...options,
+        })
+        // Sequelize automatically adds users property to checker model through association
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        if (checker.users?.some((u) => u.email === collaboratorEmail))
```
lamkeewei, line 286: If we are using the `checker` returned by `findAndCheckAuth`, then it should also be included in the same transaction. Same for the other methods.

src/client/components/dashboard/SettingsModal.tsx
rlyss: Add collaborator button is now disabled when email input field is empty. Also sets isLoading when button is clicked.

src/client/components/dashboard/CreateNewModal.tsx
```diff
@@ -128,7 +128,7 @@ export const CreateNewModal: FC<CreateNewModalProps> = ({ onClose }) => {
     createChecker.mutate({
       ...base,
       ...data,
-      isActive: false,
+      isActive: true,
```
lamkeewei: This should be enforced in the backend publish endpoint, instead of relying on the client passing `isActive` to `true`. In addition, the expected behaviour we want is such that a checker is only set to `active` on the first publish. If it has already been previously published but set to inactive, we do not want to automatically turn it to active.

src/server/database/migrations/20210706024645-modify-checkers-add-isactive-column.js
```diff
@@ -2,19 +2,10 @@
...
-    })
-
-    const query =
-      'UPDATE checkers ' +
-      'SET "isActive" = EXISTS( SELECT 1 FROM checkers as c1, "publishedCheckers" as pc1 ' +
-      'WHERE c1.id = pc1."checkerId" ' +
-      'AND checkers.id = pc1."checkerId")'
-    await queryInterface.sequelize.query(query, {
-      raw: true,
+      defaultValue: true,
```
lamkeewei: Hmm, wouldn't this result in a situation where checkers that are created are immediately `active`? We only want it to be active on first publish.

src/server/checker/CheckerService.ts
```diff
@@ -164,6 +190,114 @@ export class CheckerService {
...
+          throw new Error('User is already a collaborator')
+        if (collaboratorUser) {
+          // Sequelize automatically adds addChecker method to user model through association
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
+          await collaboratorUser.addChecker(
```
lamkeewei: Let's change this to manually creating a `UserToChecker` instance. I had a weird situation where I was unable to add a collaborator for a newly created user. Might be related to the transaction issue we faced previously.

lamkeewei (approved):
lgtm! Tested the following and works as expected:
- All published checkers should remain published and active
- Turning a checker to inactive would hide it from the public
- Publishing an unpublished checker should also make it active immediately
- After the first publish, if checker has been set to inactive afterwards, publishing should not activate the checker automatically
- Newly created checkers start off as inactive
- Checkers without published checkers cannot be activated
- Share link redirect button should be disabled if checker inactive
- Error toast if added collaborator is not a user yet
- User cannot add themselves as a collaborator
- Returns to dashboard if collaborator deletes themselves
- Added collaborators should be able to access and edit checker
- Removed collaborators should no longer be able to access checker
- Inactive checker should be displayed on the dashboard

PR #980, opened 2021-08-30
fix(masthead): shorten "how to identify" line

LoneRifle:
## Problem and Solution
The "how to identify" line as proposed by our designers used language 
as defined by this change. Remove the extraneous " it?" bit to align with
other places where this is found, eg, sites built on Isomer


lamkeewei (approved):
lgtm! Thanks for the catch @LoneRifle

PR #1102, opened 2021-10-04
feat: remove extra whoami call for public checker page

lamkeewei:
## Problem

Public checker page should not require a call to `/whoami`.

## Solution

**Features**:
- Remove `/whoami` call on public checker page

**Others**
- Manually excluded frontend dependencies in `serverless.yml`. This is not ideal and should be tackled through addressing issues #1103 

## Tests
- [ ] Verify that public checker page still works as expected and `/whoami` is not called
- [ ] Verify that logged in user should still get redirected to `/dashboard` when navigating to landing page
- [ ] Non-existent URLs should still get redirected to the landing page
- [ ] Navigation through the dashboard still still work as expected

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -31,6 +31,13 @@ provider:
       Resource:
         '*'
 
+package:
+  exclude:
+    - node_modules/@chakra-ui/**
+    - node_modules/@emotion/**
+    - node_modules/@loadable/**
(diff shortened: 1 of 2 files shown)


LoneRifle (approved):
eh, lgtm

PR #1258, opened 2022-02-28
fix: fix OGP Footer to show year dynamically and text in black

zxt-tzx:
## Problem

Right now, the footer has two problems:
- The date shown is outdated
- The text is in gray (#EEEEEE), which is hard to see against a white background

## Solution
- Modelling after AskGov, I have installed the `date-fns-tz` library to dynamically generate the date in the footer
- Have changed the text to white

As part of installing a new dependency, it seems like the `package-lock.json` has been significantly changed; I'm not sure whether there's any issues with that (what I did was to run `npm ci` per the instructions in the README, then `npm i date-fns-tz`. 

More generally, how do I know whether I'm updating `package-lock.json` correctly when adding a new dependency? I have encountered situations where I clone a new codebase, do `npm i` and have a changed `package-lock.json` — I often wonder if it's because I did sth wrong during the setup or if it's because the previous `package-lock.json` is outdated hmm.

## Before & After Screenshots

**BEFORE**:
SCR-20220228-ic4

**AFTER**:
SCR-20220228-ibz


prestonlimlianjie (approved):
lgtm

PR #1271, opened 2022-04-14
feat: improve validation for OtpForm and LoginForm

zxt-tzx:
While working on https://github.com/opengovsg/CheckWho/pull/251, thought to transpose the code changes to checkfirst.

Advantages:
- More stringent frontend validation for email; must be *@*.gov.sg email, and not just any email
- Error case looks more aesthetic using react-hook-form and Chakra UI's FormControl, compared to default HTML

BEFORE:
SCR-20220414-uqh
image

AFTER:
SCR-20220414-st9

### DIFF
--- a/src/client/components/LoginForm.tsx
+++ b/src/client/components/LoginForm.tsx
@@ -71,35 +71,38 @@ export const LoginForm: FC<LoginFormProps> = ({ email, onLogin }) => {
     )
   }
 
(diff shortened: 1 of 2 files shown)


LoneRifle (changes requested):
@tan-zx - given that we allow other quasi-government institutions to use CheckFirst (most notably the medical clusters), could you take the trouble to rework this PR so that we only take on board the OTP validation? Thanks =)

PR #1277, opened 2022-04-21
fix: delete the correct condition instead of the last one #1266

### LINKED ISSUE #1266: bug in deleting conditions
when user try to delete one of the many conditions, only the last condition is removed. hence, they hve to resort to replacing the conditions instead of just deleting it.

https://user-images.githubusercontent.com/23736580/159880793-c7587d8e-7593-4016-a7a3-325f449aa65c.mov

user should be able to delete any of the conditions in between.

aniruddha-adhikary:
## Problem

Closes #1266

## Solution

I just used the passed the ID of a Condition as the key, instead of the index.

**Bug Fixes**:

- Deletes the correct entry from Conditions, when the corresponding trash icon is clicked.

## Before & After Screenshots

**BEFORE**:
before

**AFTER**:
after

## Tests

_What tests should be run to confirm functionality?_

## Deploy Notes

_Notes regarding deployment of the contained body of work. These should note any
new dependencies, new scripts, etc._

- `env var` : env var details

**New scripts**:

- `script` : script details

**New dependencies**:

- `dependency` : dependency details

**New dev dependencies**:

- `dependency` : dependency details

### DIFF
--- a/src/client/components/builder/logic/ConditionalResult.tsx
+++ b/src/client/components/builder/logic/ConditionalResult.tsx
@@ -205,7 +205,7 @@ const InputComponent: OperationFieldComponent = ({
         <Box sx={styles.deleteSpacer} />
       </HStack>
       {conditions.map((cond, i) => (
-        <HStack key={i} sx={styles.inputContainer}>
+        <HStack key={cond.id} sx={styles.inputContainer}>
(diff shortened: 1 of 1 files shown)


LoneRifle (approved):
lgtm

PR #1340, opened 2022-11-10
fix(if): support Unit types

LoneRifle:
## Problem

Dates cannot be used as arguments for `ifelse` and `countif` functions

## Solution

Allow for Unit data types in `evaluator` expressions for those functions


lamkeewei (approved):
lgtm

PR #1487, opened 2024-09-16
docs(readme): update 'table not found error fix'

frederike-ramin:
## Problem

In the readme, you provide a very handy possibility to fix the "table not found" error. Only the path to the file to adapt is out-of-date.

## Solution

Used the correct file and updated the code to insert to avoid syntax (`)`) and formatting (` `) errors.

## Tests

/

## Deploy Notes
/


LoneRifle (approved):
lgtm

